PR #17, opened 2020-03-03
Sprint1/issue#3

nbaehler:
I open the PR in order to be able to work on it more efficiently. This is probably not the final version but in this way we have a nice thread of discussion for improvements


app/src/main/java/ch/epfl/qedit/backend/Backend.java
```diff
@@ -0,0 +1,13 @@
+package ch.epfl.qedit.backend;
+
+import ch.epfl.qedit.frontendBackendInterface.FallibleEntity;
+import ch.epfl.qedit.frontendBackendInterface.Request;
+import ch.epfl.qedit.frontendBackendInterface.Response;
+
+public abstract class Backend extends FallibleEntity {
```
brunnerant: We could consider making it an interface maybe ?

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/FallibleEntity.java
```diff
@@ -0,0 +1,17 @@
+package ch.epfl.qedit.frontendBackendInterface;
+
+public abstract class FallibleEntity {
```
brunnerant: Interface or abstract class ?

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Status.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public boolean isOk() {
+        return ok;
+    }
+
+    public String getMessage() {
+        return message;
+    }
+
+    public void error(String message) {
```
brunnerant: Maybe we could make it a static factory method ?

brunnerant: We could build statuses like this
`Status status = Status.ok();`
`Status status = Status.error("Something went wrong");`

app/src/main/java/ch/epfl/qedit/backend/Backend.java
cjordan7: Why ? Do you want to create dozens of backends ? I don't even understand why he created an abstract class for the backend

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
```diff
@@ -0,0 +1,15 @@
+package ch.epfl.qedit.backend;
+
+import ch.epfl.qedit.frontendBackendInterface.Request;
+import ch.epfl.qedit.frontendBackendInterface.Response;
+
+public class Firebase extends Backend {
+    public Firebase() {
```
cjordan7: Wasn't it supposed to be my job to do that ?

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.qedit.frontendBackendInterface;
+
+public abstract class Request extends FallibleEntity {
+    public Request() {
```
cjordan7: Why do you always create an abstract class before creating a class ?

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/RequestLogin.java
```diff
@@ -0,0 +1,19 @@
+package ch.epfl.qedit.frontendBackendInterface;
+
+public class RequestLogin extends Request {
+    private String id;
+    private String password;
+
+    public RequestLogin(String id, String password) {
+        this.id = id;
+        this.password = password;
```
cjordan7: I wonder if for the moment if creating the classes RequestLogin and ResponseLogin aren't a little overkill.

app/src/main/java/ch/epfl/qedit/backend/Backend.java
nbaehler: Because we want to make the application independent of the backend running behind it. We abstract it away. This makes switching the backend easy. It's not our goal to "create dozens of backends".

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
nbaehler: Yes but as you can see this class is pretty much empty. It's just a stub, a placeholder, for me that made it a bit easier for me to actually do something. When merging your code will be put inside this class.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
nbaehler: Some of them will be turned into interface like @brunnerant suggested ... But there will be more types of requests than just the request of login. So the abstract class or the interface allows the group common code/properties for all request. Keep in mind, this is just the first version of a structure of classes and interfaces, it will be extended in the future.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/RequestLogin.java
nbaehler: For now it's overkill as we do not have other requests ...

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/FallibleEntity.java
nbaehler: I'll turn it into a interface.

app/src/main/java/ch/epfl/qedit/backend/Backend.java
cjordan7: We can create a backend class and then hide all the different functions of our server, namely Firebase in it. Something like an adapter, which I did it in a `Server class`. But you it is not what you seem to want to do. You don't even hide firebase later because you created a class in the open for all our application to see.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
cjordan7: So what's going to happen ? We want a request for a name and you will create a RequestName ? You want a request for a quiz you create a class RequestQuiz ? You want a request for a banana you create a class RequestBanana ? It will become unmanageable and a big mess.

In Firebase side it is quite simple to login, usually just a function for example with password and email just, firebase.signinwithemailpassword(email, password). You seem to add a lot of layer of complexity from our side.

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
cjordan7: > Because we want to make the application independent of the backend running behind it. We abstract it away.

But the goal is to hide Firebase and not showing that we use Firebase. What about abstracting it now ?

> When merging, your code will be put inside this class.

You're kidding ? The goal is to hide Firebase behind an adapter (`Server` in the already merged code). Not show it to our application that we used Firebase.

Delete that awful class.

app/src/main/java/ch/epfl/qedit/backend/Backend.java
nbaehler: Well the code for the real backend has to be somewhere ... By hiding I mean that we only use the abstractions in the apps code except for the code that implements the real backend. I don't see how you want to circumvent having that code somewhere. And I if you'd open a PR we could actually see what your doing.

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
nbaehler: No need to get toxic. "Backend" plays to role of an adapter or at least it will once we have your code.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
nbaehler: Yes, I already implemented the RequestBanana.

nbaehler: Until now I implemented more or less what we have discussed last Friday. Maye it would be better to simply add methods to Backend or Server instead of creating the different classes of Requests and Responses.

app/src/main/java/ch/epfl/qedit/backend/Backend.java
cjordan7: Oh yeah. I see now.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
cjordan7: I think we will have to discuss with the other more in detail about that.

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
cjordan7: Or delete that class and then make the `Server` class that I have already created inherit from Backend. No need for anyone else in the app to be aware that Firebase exists.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Request.java
nathan96g: what i propose for this pull request is to wait until the sprint review, and we will decide the correct architecture friday morning. We will get the work redistributed then because we will need to put the back end front end and interface working together for the next sprint

app/src/main/java/ch/epfl/qedit/backend/Firebase.java
nbaehler: Yes that is a good option. Firebase is or was the equivalent to Server.

app/src/main/java/ch/epfl/qedit/frontendBackendInterface/Status.java
nbaehler: I implemented this behavior.

nathan96g:
This classes are not needed anymore

PR #18, opened 2020-03-03
created users classes and quiz class

alexisc99:
Dites moi ce que vous en pensez :)


app/src/main/java/ch/epfl/qedit/model/QuizCategorie.java
```diff
@@ -0,0 +1,6 @@
+package ch.epfl.qedit.model;
+
+public enum QuizCategorie {
```
brunnerant: Maybe, with a list of questions, we don't really need this enum

app/src/main/java/ch/epfl/qedit/model/users/Administrator.java
```diff
@@ -0,0 +1,20 @@
+package ch.epfl.qedit.model.users;
+
+import java.util.ArrayList;
+import java.util.Date;
```
brunnerant: We should use `Calendar` instead of `Date`, because the documentation says it is deprecated...

app/src/main/java/ch/epfl/qedit/model/users/Participant.java
```diff
@@ -0,0 +1,41 @@
...
+    private Quiz quiz;
+    private Date startTime;
+
+
+    public Participant(String firstName, String lastName, int userId, String language) {
+        super(firstName, lastName, userId, language);
+    }
+
+    //allows participant to select a quiz or be assigned one
+    public boolean getQuiz(Quiz quiz){
```
brunnerant: Do we really need the methods `setQuizStartTime`, `createQuiz`, `editQuiz`, `getQuiz`, .... ?
I feel like it creates a lot of coupling between `User` and `Quiz`, which are conceptually not so related. IMO, it will make the code harder to test and maintain.
Furthermore, I think that the "model" part of the MVVM shouldn't care too much about behavior, but rather just hold the data.
What do you think ?

alexisc99: As it is still unclear to me how the editor and the quiz will interact I added these methods as "might be useful later" we'll have to talk about these interactions. Other than that, I agree with the fact that the "model" shouldn't care too much about behavior.

app/src/main/java/ch/epfl/qedit/model/users/Administrator.java
alexisc99: Ok I'll change this

app/src/main/java/ch/epfl/qedit/model/QuizCategorie.java
alexisc99: Yep

PR #19, opened 2020-03-04
Create a LoginActivity and a ViewRoleActivity

Anthony-Iozzia:
This is a very first version. I will clean it.
I generated a default LoginActivity and I translated it in French.


Anthony-Iozzia:
I created a LoginActivity that redirects to a ViewRoleActivity whenever the login is successful.
Since I don't know yet the way to access the user data with the backend, the user "admin" will set the role as "administrator" and all other usernames will set the role as "participant".
In the next steps, I will link the authentication with the backend.
Codeclimate and travis do not pass, help :/

Anthony-Iozzia:
Codeclimate says that the method onCreate in class LoginActivity has 94 lines of code (exceeds 25 allowed) and that I should consider refactoring. But this is a predefined method created by Android Studio that I just completed with what I wanted. Refactoring it would complicate it. What do you think?

app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="ch.epfl.qedit">
+    package="ch.epfl.qedit" >
```
nbaehler: Do we need these spaces in this file?

nbaehler: Do we need these spaces in this file?

Anthony-Iozzia:
I removed all the auto-generated classes.
The coverage on the classes that I did is 100%.
I think we are good!
Can I merge?

PR #20, opened 2020-03-04
Create a `Fragment` to handle answering single field questions

nathan96g:
New Activity for a Quiz implemented
New Fragment for Question and Answer implemented
Manifest.xml updated: you can now choose the launch the app on QuizActivity for testing


app/src/main/java/ch/epfl/qedit/view/question/AnswerFragment.java
```diff
@@ -0,0 +1,31 @@
...
+import android.os.Bundle;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import ch.epfl.qedit.R;
+
+public class AnswerFragment extends Fragment {
```
brunnerant: This class is for later, right ?

brunnerant:
I removed the `if (savedInstanceState == null) ...` because we don't save the state of the application yet. I also removed an empty class which was lowering our coverage. I think that it should be ready to merge now...

app/src/main/java/ch/epfl/qedit/view/question/AnswerFragment.java
nathan96g: yes

PR #21, opened 2020-03-04
Firebase wrapper implementation (simple and start) + tests

cjordan7:


cjordan7:
Codeclimate is broken... The two locations doesn't do the same thing...

brunnerant:
Ce que j'ai fait c'est que j'ai indiqué à codeclimate que c'était pas du code dupliqué. Du coup les status checks passent mtn.

brunnerant:
No, he did it right, because he forked the repo, which achieves the same result...

app/src/main/java/ch/epfl/qedit/model/Server.java
```diff
@@ -0,0 +1,115 @@
...
+
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.auth.AuthResult;
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.FirebaseUser;
+
+public class Server {
+    private boolean isSignInSuccessful = false;
```
brunnerant: Why exactly do we need `isCreateSuccessful` and `isSignInSuccessful`, since the methods `signIn` and `createUser` already return a boolean flag ?

app/src/main/java/ch/epfl/qedit/model/Server.java
```diff
@@ -0,0 +1,115 @@
...
+    private boolean isCreateSuccessful = false;
+    private FirebaseAuth firebaseAuth = null;
```
brunnerant: Looking at the documentation, it seems that we should use `FirebaseAuth.getInstance()` to get a new instance of the authentication system.

brunnerant:
Otherwise, LGTM.

app/src/main/java/ch/epfl/qedit/model/Server.java
cjordan7: Because it is `task.isSuccessful()` that will say if either `signIn` or `createUser` will be successful and we need to access it outside of the innerClass, so I created those Boolean variables `isCreateSuccessful` and `isSignInSuccessful`.

But I just realized that I can delete them and use the method `.isSuccessful()` on both `createUserWithEmailAndPassword` and `signInWithEmailAndPassword`. I could delete a few methods as well. I think I will keep the `onComplete` method as well. It could be useful maybe later.

Anthony-Iozzia:
> LGTM.

I thought it meant "Let's get this merged" but I googled it and it means "Looks good to me".
Everyday we learn new stuff!

app/src/main/java/ch/epfl/qedit/model/Server.java
cjordan7: That is exactly what happens in `initServer()`

app/src/main/java/ch/epfl/qedit/model/Server.java
cjordan7: But I think for now I will do it this way and for later versions I will study how to get the error codes to have something better in the long run.

app/src/main/java/ch/epfl/qedit/model/Server.java
brunnerant: Maybe I'm wrong but look here:
https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth

brunnerant: My bad, I didn't see ;)

app/src/main/java/ch/epfl/qedit/model/Server.java
```diff
@@ -0,0 +1,92 @@
...
+
+    //    public void signUpNewUser(User user) {
+    //    }
+
+    //    public void child() {
+    //
+    //    }
+
+    // TODO change with user
+    public boolean signIn(String email, String password) {
```
brunnerant: Thanks for the changes ! One last question: what exactly does `isOut` mean ?  If I understand correctly, it reflects whether a user is logged in, right ? Then, we should set it to true if the sign in was successful, shouldn't we ?

app/src/test/java/ch/epfl/qedit/model/ServerTest.java
```diff
@@ -0,0 +1,28 @@
...
+        server.signOut();
+        assertTrue(server.isOut());
+    }
+
+    @Test
+    public void test2() {
+        Server server = Server.getInstance();
+        server.initServer();
+
+        assertFalse(server.createUser("", ""));
```
brunnerant: Shouldn't the server be able to create the user ?

app/src/main/java/ch/epfl/qedit/model/Server.java
```diff
@@ -0,0 +1,92 @@
+package ch.epfl.qedit.model;
+
+import androidx.annotation.NonNull;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.auth.AuthResult;
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.FirebaseUser;
+
+public class Server {
```
brunnerant: I am wondering if this class should really be called `Server`, that's maybe to generic. And it probably shouldn't be part of the `model` package, but rather in a `backend` package, or the like.

app/src/test/java/ch/epfl/qedit/model/ServerTest.java
cjordan7: That's what createUser do. Or at least in this case it sends an email and password. But later we could add more things but we didn't spoke yet about what kind of things we were going to store on the server and how we can access it. 

Firebase let us access the database by email, google, facebook, and some other methods.

app/src/main/java/ch/epfl/qedit/model/Server.java
cjordan7: I used that name to hide firebase because it really is just a wrapper, an adapter for it.

app/src/main/java/ch/epfl/qedit/model/Server.java
cjordan7: Yeah exatcly

app/src/test/java/ch/epfl/qedit/model/ServerTest.java
brunnerant: Then, why is it `assertFalse` and not `assertTrue` ? Creating a new user should succeed, right ?

app/src/main/java/ch/epfl/qedit/model/Server.java
brunnerant: Ok, the name is fine, I agree. But I think we need to move it to another package, becaues it really doesn't belong to the model part of MVVM.

PR #28, opened 2020-03-08
Link frontend and backend for login

brunnerant:
This pull requests aims at creating a fully usable login activity that connects to the Firebase backend to authenticate users in the application.


app/src/androidTest/java/ch/epfl/qedit/QuizActivityTest.java
```diff
@@ -5,14 +5,14 @@
 import static androidx.test.espresso.matcher.ViewMatchers.withId;
 import static androidx.test.espresso.matcher.ViewMatchers.withText;
 
-import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner;
```
nbaehler: What is the advantage of using this instead of the old one?

nathan96g: this is probably useful to Antoine's tests, i will ask him tomorrow

app/src/main/java/ch/epfl/qedit/view/QuizActivity.java
```diff
@@ -23,7 +23,7 @@ protected void onCreate(Bundle savedInstanceState) {
         // Pass the first question as argument to the new QuestionFragment
         Bundle bundle = new Bundle();
         bundle.putSerializable("q0", question);
-        QuestionFragment frag = QuestionFragment.newInstance();
+        QuestionFragment frag = new QuestionFragment();
```
nbaehler: The newInstance() seems to be a common pattern especially when creating classes with Android Studio. But it is indeed not very useful ...

app/src/androidTest/java/ch/epfl/qedit/QuizActivityTest.java
nbaehler: Ok, I just couldn't find the answer right away by googling ...

brunnerant:
Note to the SDP staff:
Nathan and brunnerant both worked 8 hours on this PR, but we couldn't add both tags ;).

PR #29, opened 2020-03-11
created empty QuizOverviewFragment

Anthony-Iozzia:
Started QuizOverviewFragment

PR #30, opened 2020-03-11
Sprint2/issue#24 and Sprint2/issue#27

nbaehler:
I used the Visitor pattern that was already implemented for getting the actual type of the AnswerFormat. It is then passed on to the AnswerFragment. For now I didn't write any tests.


app/src/main/java/ch/epfl/qedit/view/question/QuestionFragment.java
```diff
@@ -34,17 +37,38 @@ public View onCreateView(
...
     @Override
     public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
     }
+
+    // Concrete Visitor
+    private class Visitor implements AnswerFormat.Visitor {
+        @Override
+        public void visitNumberField(AnswerFormat.NumberField field) {
+            answerFormat = field;
```
nbaehler: I don't really like this way of doing it but I think that the accept() method normally returns void so I kind of had to add a private field and do it in this way.

Anthony-Iozzia:
We added my work (pull request #29) in this branch and I will continue here.

PR #31, opened 2020-03-12
SingleFieldFragment

cjordan7:


brunnerant:
Should we keep this PR, or should we directly work on a more generic answer fragment ?
I am asking this because both #31 and #32 could be backed by a more generic answer fragment.

PR #32, opened 2020-03-12
created 3by3 matrix fragment

alexisc99:
Should now take bundle arguments to create MxN matrices


app/src/main/java/ch/epfl/qedit/view/question/TableFragment.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * A simple {@link Fragment} subclass. Use the {@link TableFragment#newInstance} factory method to
+ * create an instance of this fragment.
+ */
+public class TableFragment extends Fragment {
+    private int sizeX;
+    private int sizeY;
+    private List<View> answers;
+
+    private Button buttonSubmit;
```
nbaehler: We don't need this button.

app/src/main/res/values/strings.xml
```diff
@@ -9,4 +9,8 @@
     <string name="role">"Display role"</string>
     <string name="buttonOpenQuiz">"Quiz test"</string>
     <string name="buttonLogin">"Connexion"</string>
+
+    <!-- TODO: Remove or change this placeholder text -->
```
nbaehler: We probably can delete these lines.

app/src/main/java/ch/epfl/qedit/view/question/TableFragment.java
alexisc99: Yes, I forgot to remove it. Done now.

app/src/main/res/values/strings.xml
alexisc99: Of course

PR #35, opened 2020-03-14
Change the general flow of the app and add a Home Activity

nbaehler:
For now a simply changed to order of the activities such that we get the following flow: Login -> Home -> Quiz. In the home activity I simply show what we had before in the show_role activity and I took the button of the main activity that starts the quiz and put it below these information.


app/src/main/java/ch/epfl/qedit/backend/database/FirebaseDBService.java
```diff
@@ -0,0 +1,54 @@
...
+                .addOnCompleteListener(
+                        new OnCompleteListener<DocumentSnapshot>() {
+                            @Override
+                            public void onComplete(@NonNull Task<DocumentSnapshot> task) {
+                                Response<BundledData> response;
+                                if (task.isSuccessful()) {
+                                    DocumentSnapshot document = task.getResult();
+                                    if (document.exists())
+                                        if (collection.equals("quizzes"))
+                                            response = Response.ok(getQuizFromDocument(document));
```
brunnerant: Just as a helper: I created the `BundledData` class with firestore in mind. This means that the data from the document can directly be extracted by doing `new BundledData(document.getData())`.
It is then the responsibility of the caller of the database query to transform the data into a Java type. Therefore you don't need to interpret the data in the database implementation, but rather just return it.

brunnerant: To clarify, it means that you have to write:
`response = Response.ok(new BundledData(document.getData());`
instead of what you wrote.

Anthony-Iozzia:
What is the name of this branch?

nbaehler:
> What is the name of this branch?

sprint3/issue#33

app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
```diff
@@ -0,0 +1,78 @@
...
+            return;
+
+        Question question = quiz.getQuestions().get(index);
+
+        // We have to change the question title and text
+        String questionTitleStr = "Question " + (index + 1) + ": " + question.getTitle();
+        questionTitle.setText(questionTitleStr);
+        questionDisplay.setText(question.getText());
+
+        // And dynamically instatiate the answer form
```
Anthony-Iozzia: Typo instantiate

app/src/main/java/ch/epfl/qedit/viewmodel/QuizViewModel.java
```diff
@@ -0,0 +1,56 @@
...
+import ch.epfl.qedit.util.BundledData;
+import ch.epfl.qedit.util.Callback;
+import ch.epfl.qedit.util.Response;
+
+public class QuizViewModel extends ViewModel {
+    public enum Status {
+        NotLoaded,
+        Loading,
+        Loaded,
+        CouldntLoad
```
Anthony-Iozzia: "CouldNotLoad" would be better, no?

app/src/test/java/ch/epfl/qedit/model/UserTest.java
```diff
@@ -19,4 +21,30 @@ public void dummyTestUser() {
...
+    public void quizzesTest() {
+        User user = new User("John", "Doe", User.Role.Administrator);
+        user.addQuiz("q0", "First Quiz");
+        user.addQuiz("q1", "Second Quiz");
+        assertThat(user.getQuizzesKey(), hasItems("q0", "q1"));
+        assertThat(user.getQuizzesTitle(), hasItems("First Quiz", "Second Quiz"));
+    }
+
+    @Test
+    public void addQuizzesAddlTest() {
```
Anthony-Iozzia: Addl?

app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
nbaehler: Fixed

app/src/main/java/ch/epfl/qedit/viewmodel/QuizViewModel.java
nbaehler: I think that CouldntLoad stands for "Couldn't Load" but I prefer your version as well. I'm going to change it.

app/src/test/java/ch/epfl/qedit/model/UserTest.java
nbaehler: I honestly do not know what the Addl stands for but following the naming of the other tests I'm going to replace it by: addQuizzesTest.

Issue #34, opened 2020-03-14
Finish the question fragment so that it is usable by the end of sprint 3


brunnerant:
This is not a new feature per se, but rather combining, cleaning up, and testing more thoroughly the code from issue #30.

PR #36, opened 2020-03-14
Sprint3/issue#34

brunnerant:


brunnerant:
Here is an overview of what I added in this PR:
- I created a `QuizViewModel` for the quiz activity and its fragments, so that they can all share the data, and automatically update when it comes from the database.
- I created a `DatabaseService` interface to abstract away the Firestore database. For now it is only implemented by `MockDBService`, which, as its name implies, is a mock database. Note that the mock database is already asynchronous, so that I could test how the view would handle asynchronous requests.
- Related to the previous point, I created a `DataBundle` class that represents a piece of data from the database. I also created a `Bundlabe` interface for objects that can be transformed to a bundle, and therefore sent to our database.

app/src/main/java/ch/epfl/qedit/backend/auth/AuthenticationService.java
```diff
@@ -0,0 +1,23 @@
...
+import ch.epfl.qedit.model.User;
+import ch.epfl.qedit.util.Callback;
+import ch.epfl.qedit.util.Response;
+
+/**
+ * Represents an authentication service, that is a service to which login requests can be sent. It
+ * handles asynchronous responses by using a callback.
+ */
+public interface AuthenticationService {
+    int CONNECTION_ERROR = 1;
```
cjordan7: Maybe we could use an enum for those ERRORs. I have seen as well that you use them in other classes. Same things for the WRONG_STUFFs

app/src/main/java/ch/epfl/qedit/util/Response.java
```diff
@@ -0,0 +1,38 @@
+package ch.epfl.qedit.util;
+
+/**
+ * Represents a response that a certain service can send after a request.
+ *
+ * @param <T> the type of the data when the response is successful
+ */
+public class Response<T> {
+    public static final int NO_ERROR = 0;
```
cjordan7: We could use an enum here, as well. the same as above

app/src/main/java/ch/epfl/qedit/viewmodel/QuizViewModel.java
```diff
@@ -0,0 +1,56 @@
...
+import ch.epfl.qedit.util.BundledData;
+import ch.epfl.qedit.util.Callback;
+import ch.epfl.qedit.util.Response;
+
+public class QuizViewModel extends ViewModel {
+    public enum Status {
+        NotLoaded,
+        Loading,
+        Loaded,
+        CouldntLoad
```
cjordan7: But here, you create the enum... ?

app/src/main/java/ch/epfl/qedit/backend/auth/AuthenticationService.java
brunnerant: Yes, initially that's what I did. The problem is that different services can have different errors, so making an enum would mean that all those errors would be grouped in a single place, which I think doesn't make sense.

app/src/main/java/ch/epfl/qedit/viewmodel/QuizViewModel.java
brunnerant: That enum is not related to a service (auth or database) but rather to the state of the quiz activity. It is used by fragments to react when the data is loading from the database (progress bar), and to print a toast in case of error (couldn't load)

app/src/main/java/ch/epfl/qedit/backend/auth/AuthenticationService.java
nathan96g: Yes i first thinked like you Cosme and then i understood why antoine did it that way.

cjordan7: I agree different services can have different errors but to me it seems that AuthentificationService and Response do have the same kinds. 

By the same way using int doesn't make sense to me.

nathan96g: Responce is generic and is used by AuthService and DatabaseService, but we could use it in different context (other than backend related stuff) i think

app/src/main/java/ch/epfl/qedit/view/question/QuestionFragment.java
```diff
@@ -41,42 +27,53 @@ public View onCreateView(
...
     }
 
-    @Override
-    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
-        super.onActivityCreated(savedInstanceState);
-    }
+    /** Handles the transition from one question to another */
+    private void onQuestionChanged(Quiz quiz, Integer index) {
+        if (index == null || quiz == null || index < 0 || index >= quiz.getQuestions().size())
+            return;
```
cjordan7: Why do you test your parameters here and sometimes you don't ?

app/src/main/java/ch/epfl/qedit/backend/auth/AuthenticationService.java
nathan96g: Do you receive this comments?

PR #38, opened 2020-03-16
Sprint3/issue#37 Toolbar

alexisc99:


app/src/main/java/ch/epfl/qedit/model/Question.java
```diff
@@ -12,6 +12,7 @@
 
     private final String text;
 
+    private int index;
```
brunnerant: Do we need this field ? I think we removed it in a previous PR, and now we add it again...

alexisc99: Actually we don't. I re-added it before seeing the liveData Integer structure you implemented and forgot to remove it, thanks !

brunnerant:
LGTM. Just don't forget to remove the unnecessary commented code before merging.

app/src/main/res/values/styles.xml
```diff
@@ -1,7 +1,7 @@
 <resources>
 
     <!-- Base application theme. -->
-    <style name="AppTheme" parent="Theme.AppCompat.Light.DarkActionBar">
+    <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
```
nathan96g: You cannot really forced all the application to not have the Action Bar, was it the only solution ? Actually what was the problem with adding things to the existing action bar ?

alexisc99: Corrected this ;)

PR #39, opened 2020-03-16
Sprint3/issue#14 MatrixFormat and MatrixFragment done

cjordan7:
There are still comments and tests to be done, though.


app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
```diff
@@ -0,0 +1,107 @@
...
+import ch.epfl.qedit.model.MatrixFormat;
+import java.util.ArrayList;
+
+public class MatrixFragment extends Fragment {
+    private TableLayout tableLayout;
+    private boolean hasDecimal = true;
+    private boolean hasSign = true;
+
+    private ArrayList<TableRow> tableRow = new ArrayList<>();
+    private ArrayList<ArrayList<EditText>> arrayButtons = new ArrayList<>();;
```
nbaehler: Typo

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        matrixFormat = (MatrixFormat) getArguments().getSerializable("m0");
+        setValues();
+    }
+
+    // Used to get the context
+    @Override
```
nbaehler: Why do you exactly need these two methods? You don't seem to need the context.

app/src/main/res/layout/answers_table.xml
```diff
@@ -0,0 +1,47 @@
...
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content">
+
+            <TableLayout
+                android:id="@+id/answersTable"
+                android:layout_width="wrap_content"
+                android:layout_height="match_parent">
+
+                <TableRow
+                    android:layout_width="match_parent"
```
brunnerant: Why do we need those table rows if we are already instantiating them dynamically in MatrixFragment ?

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
```diff
@@ -0,0 +1,74 @@
...
+    private Visitor visitor;
+    private boolean hasDecimal = true;
+    private boolean hasSign = true;
+
+    private int tableRowsNumber = 1;
+    private int tableColumnsNumber = 1;
+    private int maxCharacters = 3;
+    private String hintString = hint();
+
+    public MatrixFormat(
```
brunnerant: I think we should move this class as an inner class of `AnswerFormat` for better encapsulation.

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
```diff
@@ -0,0 +1,107 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+import ch.epfl.qedit.R;
+import ch.epfl.qedit.model.MatrixFormat;
+import java.util.ArrayList;
+
+public class MatrixFragment extends Fragment {
+    private TableLayout tableLayout;
+    private boolean hasDecimal = true;
```
brunnerant: Do we need `hasDecimal`, `hasSign`, `tableRowNumbers`, ... as private fields, since `matrixFormat` already contains that information ?

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
```diff
@@ -0,0 +1,107 @@
...
+                arrayButtons.get(i).add(editText);
+                tableRow.get(i).addView(editText);
+            }
+            tableLayout.addView(tableRow.get(i));
+        }
+
+        return view;
+    }
+
+    private void setValues() {
```
brunnerant: We should remove this method I think, because it is useless to duplicate data around. It will just worsen the maintainability and readability of our code ;).

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public static MatrixFormat createSingleField(
+            boolean hasDecimal, boolean hasSign, int maxCharacters) {
+        return new MatrixFormat(1, 1, hasDecimal, hasSign, maxCharacters);
+    }
+
+    @Override
+    public void accept(Visitor visitor) {
+        this.visitor = visitor;
```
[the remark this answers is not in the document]
brunnerant: And you have to add that method to our visitor.

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
cjordan7: Typo on what ? I may be blind but I really don't see the typo

nbaehler: Two semicolons, that’s it ;)

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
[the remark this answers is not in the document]
cjordan7: Oh yes

app/src/main/res/layout/answers_table.xml
cjordan7: I just dragged a table layout in android studio and it automatically added them I suppose

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
cjordan7: I thought about that and I added those because I wanted the code to be clearer and cleaner than having in some places

`editText.setKeyListener(DigitsKeyListener.getInstance(matrixFormat.hasSign, matrixFormat.hasDecimal));
`
I think here using only `hasSign` and `hasDecimal` is cleaner

`editText.setFilters(new InputFilter[] {new InputFilter.LengthFilter(matrixFormat.maxCharacters)});`

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
cjordan7: I agree with the maintainability part but I don't agree with the readability part.

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
cjordan7: I'm not sure it is a good idea to create inner classes with all formats. The more formats we have and the less readable the AnswerFormat class would be.

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
cjordan7: Yeah you are right. I thought I needed them but in the end I don't.

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public int getTableRowsNumber() {
+        return tableRowsNumber;
+    }
+
+    public int getTableColumnsNumberNumber() {
+        return tableColumnsNumber;
+    }
+
+    public boolean getHasDecimal() {
```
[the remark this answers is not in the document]
cjordan7: Sure, but I don't see how it would be clearer.

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
brunnerant: Yeah, that's right as well, so I think it is fine like this

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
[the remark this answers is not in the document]
brunnerant: I know I'm picky ;), but chaining two verbs sounds a bit weird IMO. Get and Has kind of clash together. That's a minor issue, I just noted it because I care about details.

app/src/main/res/layout/answers_table.xml
brunnerant: Ok, I think it's cleaner, thank you !

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
brunnerant: I agree with you, writing `getXX` instead of `XX` is annoying. But having those fields in the fragment is a waste of space. It also means that extra code needs to be run to "import" those attributes, when you call `setValues`.
IMO the cons outweigh the pros, but that's not such a big issue in the end.

app/src/main/java/ch/epfl/qedit/model/MatrixFormat.java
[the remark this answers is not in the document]
cjordan7: Yeah, it is fine with. I've changed it

cjordan7:
I still have to write the tests of the Fragment

brunnerant:
LGTM. Just write that last test, and we can merge ;) .

brunnerant:
Sorry, didn't mean to close.

app/src/main/res/layout/answers_table.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <ScrollView
+        android:layout_width="410dp"
```
brunnerant: I don't think it's a good idea to hard code the sizes of widget if we want our app to be used on many devices.

app/src/test/java/ch/epfl/qedit/model/MatrixFormatTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void checkMatrix3x3Normal() {
+        MatrixFormat matrixFormat = MatrixFormat.createMatrix3x3();
+
+        checkAllElements(matrixFormat, true, true, 3, 3, 5, "00000");
+    }
+
+    @Test
+    public void checkMatrix3x3WithParam() {
+        MatrixFormat matrixFormat = MatrixFormat.createMatrix3x3(true, true, 5);
+        ;
```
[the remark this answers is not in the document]
cjordan7: It is done. It had its own commit.

app/src/main/res/layout/answers_table.xml
cjordan7: I agree, as I said before it was done automatically when I dragged it into the design, next time I'll be more careful

PR #43, opened 2020-03-17
Sprint3/issue#42 Sanitize token inputs

Anthony-Iozzia:


Anthony-Iozzia:
I sanitized the token inputted by the user. Now, only a 20 characters alphanumeric string is accepted.
I also added more test questions in the mock database, and the QuizOverview now displays the titles of the questions.

PR #45, opened 2020-03-18
Sprint3/issue#44 Manage multiple languages

Anthony-Iozzia:


Anthony-Iozzia:
The convention for naming a resource file is <resource type>-b+<language code>[+<country code>]
For English, I will use the language code "en" and I will not put country code, since there is en-CH, en-US and en-UK. Or do I? I think one English language is enough.

Anthony-Iozzia:
I will continue this in pr #64

Anthony-Iozzia:
See #64

PR #62, opened 2020-03-22
Sprint4/issue#48 Extract quizzes from firestore database

nathan96g:
Extract quizzes from the firestore database.
Update the code structure

PR #63, opened 2020-03-23
Sprint4/issue#50, add some simple tests for home fragments and sort the quiz list

nbaehler:


brunnerant:
If you want I can try to factorize the two similar methods by creating a super class for tests that need to connect to the database

nbaehler:
> If you want I can try to factorize the two similar methods by creating a super class for tests that need to connect to the database, so that CodeClimate stops complaining.

On one hand we actually need to prepare the same environment so maybe it makes sense to have that refactor. On the other hand it may be a bit complicated.

nbaehler:
> If you want I can try to factorize the two similar methods by creating a super class for tests that need to connect to the database, so that CodeClimate stops complaining.

I actually did what you proposed in #62, we had the same issue there. We‘ll see if it works out when merging the code.

nbaehler:
> If you want I can try to factorize the two similar methods by creating a super class for tests that need to connect to the database, so that CodeClimate stops complaining.

Finally I added a super class here as well that does what you suggested.

app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
```diff
@@ -68,6 +70,7 @@ private void startQuizActivity(String quizID) {
 
         public CustomAdapter(Context context, ArrayList<Map.Entry<String, String>> entries) {
             this.entries = entries;
+            Collections.sort(this.entries, CustomComparator);
```
nbaehler: @cjordan7 as you handled this one your PR I should probably remove all changes to HomeQuizListFragment, do you agree?

cjordan7: Yes

nbaehler: I removed the chages

PR #64, opened 2020-03-26
Sprint4/issue#44 Manage multiple languages 2

Anthony-Iozzia:
I finished the management of multiple languages in sprint3/issue#44, but there were way too much conflicts, so here I am, merging master into my branch by hand.

### DIFF
--- a/app/src/androidTest/java/ch/epfl/qedit/LoginActivityLanguageTest.java
+++ b/app/src/androidTest/java/ch/epfl/qedit/LoginActivityLanguageTest.java
@@ -1,0 +1,91 @@
+package ch.epfl.qedit;
+
+import android.content.Intent;
+
+import androidx.test.espresso.intent.rule.IntentsTestRule;
+import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Locale;
+
+import ch.epfl.qedit.util.LocaleHelper;
+import ch.epfl.qedit.view.LoginActivity;
+
+import static androidx.test.espresso.Espresso.onData;
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.action.ViewActions.click;
+import static androidx.test.espresso.action.ViewActions.closeSoftKeyboard;
+import static androidx.test.espresso.assertion.ViewAssertions.matches;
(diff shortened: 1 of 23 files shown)


Anthony-Iozzia:
Please read the wiki before reviewing ;)

Anthony-Iozzia:
master is broken
we need to fix it

app/src/main/java/ch/epfl/qedit/view/LoginActivity.java
```diff
@@ -1,41 +1,133 @@
...
+
+        /* Language selection */
+        // Create spinner (language list)
+        Spinner languageSelectionSpinner = findViewById(R.id.language_selection);
+
+        // Find app's current language position in languages list
+        String currentLanguage = Locale.getDefault().getLanguage();
+        int positionInLanguageList = 0;
+        String[] languageList = getResources().getStringArray(R.array.languages_codes);
+        boolean languageFound = false;
```
brunnerant: We should probably use `Arrays.asList(languageList).indexOf(currentLanguage)` for more conciseness ;) ...

app/src/androidTest/java/ch/epfl/qedit/LoginActivityLanguageTest.java
```diff
@@ -0,0 +1,92 @@
...
+            String languageCode,
+            String loginString,
+            String languageChangedString) {
+        String startupLanguage = Locale.getDefault().getLanguage();
+        // We don't use this variable but the call increases the coverage ;)
+        @SuppressWarnings("unused")
+        String lang = LocaleHelper.getLanguage(testRule.getActivity());
+
+        // Language position
+        int pos = -1;
```
brunnerant: How do we know that "en" has position 0 and "fr" has position 1 ? Are the languages sorted lexicographically ? I am asking that question because I don't want the test to fail because the list gets reordered.

Anthony-Iozzia: You are right, we must not reorder the language list, else it will not work anymore. I tried to compute the position of the language with <the activity rule>.getActivity().getArray(R.array.languages_list) but it does not work (cannot resolve method getArray(int)

app/src/main/java/ch/epfl/qedit/view/LoginActivity.java
Anthony-Iozzia: That's very clean, I will do this, thank you!

cjordan7:
as @brunnerant says: Let's Get This Merged.

PR #74, opened 2020-03-30
Sprint5/issue#71 Create a MultiFieldAnswerFormat 

nathan96g:
Package answer created
Adding optional text to AnswerFormat
MultiFieldAnswerFormat implemented
Parsing re-think


app/src/test/java/ch/epfl/qedit/model/MultiFieldFormatTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+import ch.epfl.qedit.model.answer.AnswerFormat;
+import ch.epfl.qedit.model.answer.MatrixFormat;
+import ch.epfl.qedit.model.answer.MultiFieldFormat;
+import org.junit.Test;
+
+public class MultiFieldFormatTest {
```
brunnerant, line 12: Maybe the field to enter the number of questions should be explicit here. But that's just a note for future PRs, I think it is fine like this for this PR.

brunnerant:
LGTM. It looks like there are a lot of renaming in this PR which will probably create some conflicts when merged. But it is good to keep the code consistent, so good job !

app/src/androidTest/java/ch/epfl/qedit/HomePopUpTest.java
```diff
@@ -1,4 +1,5 @@
-/*package ch.epfl.qedit;
```
nbaehler, line 1: Why is this entire class commented?

app/src/androidTest/java/ch/epfl/qedit/answer/MatrixFragmentTest.java
```diff
@@ -37,7 +37,7 @@ public void init() {
         DatabaseFactory.setInstance(dbService);
 
         Bundle bundle = new Bundle();
-        bundle.putSerializable(MatrixFragment.MATRIXID, MatrixFormat.createMatrix3x3());
+        bundle.putSerializable(MatrixFragment.MATRIX_ID, MatrixFormat.createMatrix3x3());
```
nbaehler, line 40: In the PR of Alexis and myself we have put a static constant ANSWER_FORMAT which is a bit more general. I'll probably replace MATRIX_ID by this constant when merging.

app/src/main/java/ch/epfl/qedit/model/answer/MatrixFormat.java
```diff
@@ -13,10 +13,15 @@
     private int tableColumnsNumber = 1;
     private int maxCharacters = 5;
     private String hintString;
-    private String id = MatrixFragment.MATRIXID;
+    private String id = MatrixFragment.MATRIX_ID;
```
nbaehler, line 16: I think that the ID is not really necessary, we removed it in our branch. Do you rely on it somewhere?

app/src/main/java/ch/epfl/qedit/view/answer/MatrixFragment.java
```diff
@@ -15,11 +15,11 @@
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 import ch.epfl.qedit.R;
-import ch.epfl.qedit.model.MatrixFormat;
+import ch.epfl.qedit.model.answer.MatrixFormat;
 import java.util.ArrayList;
 
 public class MatrixFragment extends Fragment {
-    public static final String MATRIXID = "ch.epfl.qedit.view.answer.MATRIXID";
```
nbaehler, line 22: Same as above -> Is going to be replaced by ANSWER_FORMAT.

app/src/androidTest/java/ch/epfl/qedit/HomePopUpTest.java
brunnerant: Because otherwise the coverage is above 80% -_-, just kidding. I think Cosme didn't write the tests because he knew I would in my PR....

nbaehler: Never mind, Antoine told me why.

app/src/androidTest/java/ch/epfl/qedit/answer/MatrixFragmentTest.java
Anthony-Iozzia: No problem, you can replace it!

app/src/main/java/ch/epfl/qedit/model/answer/MatrixFormat.java
Anthony-Iozzia: No, I just renamed it because it was detected as a typo and it triggered me. You can remove it!

Issue #67, opened 2020-03-27
Unify the RecyclerViews


nbaehler:

brunnerant:
The goal of this issue is to unify the recycler views that we have in the application. Currently, there is a recycler view in the home activity that allows an editor to edit the list of quizzes that he has. There is another one that allows an editor to add, remove and reorder questions within a quiz.
The goal is to merge them into one generic class so that the code is more maintainable, less redundant and the style is more consistent.

PR #75, opened 2020-03-31
Tackles issue #67

brunnerant:
See issue #67 for a description of what this PR aims at ;).


brunnerant:
The tests fail because I have not yet updated them to reflect the changes that I have made to `ListEditView`.

brunnerant:
This PR is good to go. There was a problem with the coverage of Cosme's PR. It may be related to the fact the he works on a fork of the repo.... Anyway, I refactored what he wrote and wrote some more thorough tests, so it is ok now.

brunnerant:
Maybe a few comments on what I did and what work would need to follow. I have added the list edit view to the home activity, which I have refactored into a nice class `ListEditView`. This class can be listened through a `ListEditAdapter`, so that all the events can be processed properly.
The thing that we will have to handle better in the next PRs is to dispatch the `HomeQuizListFragment` whether the user is an editor or a participant. Right now, all the users can edit the list of quizzes, which doesn't make sense.

app/src/main/java/ch/epfl/qedit/view/util/ConfirmDialog.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        Bundle args = getArguments();
+        message = args.getString("message");
+        listener = (ConfirmationListener) args.getSerializable("listener");
+
+        return new AlertDialog.Builder(getActivity())
```
nbaehler, line 55: Personally I have a trauma from getActivity() ... ;) Maybe it is not justified but I'm suggesting to use requireActivity()

app/src/main/java/ch/epfl/qedit/view/util/EditTextDialog.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        Bundle args = getArguments();
+        message = args.getString("message");
+        listener = (SubmissionListener) args.getSerializable("listener");
+
+        dialog =
+                new AlertDialog.Builder(getActivity())
```
nbaehler, line 126: RequireActivity() would possibly be safer.

app/src/main/java/ch/epfl/qedit/view/util/ConfirmDialog.java
brunnerant: Yes you are right, I will change that ;).

app/src/main/java/ch/epfl/qedit/view/util/EditTextDialog.java
brunnerant: Yes, I will also change that, that's a mistake from me -_-

brunnerant:
Yes, you are totally right. But I didn't want to make that monolitic PR that creates conflicts everywhere, so I will do it in a future PR. Here is a list of things we could (or have to) do in the following PRs:
 - Dispatching between participants, editors and administrators. What I mean by that is that a participant should have a different `HomeActivity` because he shouldn't be able to modify quizzes. An editor should have the editing capabilities. Right now, every one can edit quizzes in the home activity.
 - When an editor clicks on the edit button, he should be redirected to the `EditActivity` (or whatever activity we created). Right now, he is redirected to the `QuizActivity`, because the `EditActivity` has not been made yet.
 - When a participant clicks on a quiz, he should be redirected to the `QuizActivity`.
We could make tasks out of those missing features, for the next sprint.

nbaehler:
> Yes, you are totally right. But I didn't want to make that monolitic PR that creates conflicts everywhere, so I will do it in a future PR. Here is a list of things we could (or have to) do in the following PRs:
> 
>     * Dispatching between participants, editors and administrators. What I mean by that is that a participant should have a different `HomeActivity` because he shouldn't be able to modify quizzes. An editor should have the editing capabilities. Right now, every one can edit quizzes in the home activity.
> 
>     * When an editor clicks on the edit button, he should be redirected to the `EditActivity` (or whatever activity we created). Right now, he is redirected to the `QuizActivity`, because the `EditActivity` has not been made yet.
> 
>     * When a participant clicks on a quiz, he should be redirected to the `QuizActivity`.
>       We could make tasks out of those missing features, for the next sprint.

Yes I agree 100%.

PR #77, opened 2020-04-01
Sprint5/issue#53 Editor Mode: Create a question edition fragment

cjordan7:


app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
```diff
@@ -60,17 +64,39 @@ public String getText(Question item) {
...
-                        if (type == ListEditView.EventType.RemoveRequest)
-                            adapter.removeItem(position);
+                        switch (type) {
+                            case Select:
+                                setFragment(R.id.question_details_container);
+                                model.getFocusedQuestion().postValue(position);
+                                break;
+                            case RemoveRequest:
+                                adapter.removeItem(position);
+                                break;
```
nathan96g, line 81: Here, should we test if the FocusedQuestion is the one we remove? And if so set it to the previous question or a default value, i don't know.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuizActivity.java
```diff
@@ -2,19 +2,49 @@
...
+    // TODO: Delete when we connect the rest of the app
+    private void tempQuiz() {
+        Question question1 =
+                new Question("Bananas", "How many bananas are there on Earth?!", "matrix1x1");
+        Question question2 =
+                new Question("Apples", "How many apples are there on Earth?!", "matrix1x1");
+        Question question4 = new Question("Vectors", "Give a unit vector.", "matrix1x3");
+        Question question5 =
+                new Question("Operations", "What is the results of 1 + 10?!", "matrix1x1");
+        Question question6 = new Question("Matrices", "Fill those matrix !", "matrix3x3");
```
nathan96g, line 44: mdr, your question are so agressive with '?!' , i think it's fun that way x) no need to change it

app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
```diff
@@ -34,7 +34,14 @@ public View onCreateView(
...
                 new ViewModelProvider(requireActivity()).get(QuizViewModel.class);
-
+        System.out.println("=====================================");
+        System.out.println("=====================================");
+        System.out.println("=====================================");
+        System.out.println(model);
+        System.out.println("=====================================");
+        System.out.println("=====================================");
+        System.out.println("=====================================");
+        System.out.println("=====================================");
```
nathan96g, line 44: I suppose it's for debug purpose, don't forget to delete that before merging with master.

app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
```diff
@@ -50,9 +57,14 @@ public void onChanged(Integer index) {
...
-        if (index == null || quiz == null || index < 0 || index >= quiz.getQuestions().size())
-            return;
+        System.out.println("Heanvounsoinrvoirnv ori novi iorn eoin voinr oiren ");
+        System.out.println(quiz);
+        System.out.println(index);
 
+        if (index == null || quiz == null || index < 0 || index >= quiz.getQuestions().size()) {
+            return;
+        }
+        System.out.println("Question question = quiz.getQuestions().get(index);");
```
nathan96g, line 67: yes, same comment as above, just highlighting (it is not spelled correctly right ^^ ) so that you don't forget to delete it.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuizActivity.java
cjordan7: Well it was for code climate. I tried things so it stopped saying "Duplicated code" because in the end, more likely in a week or two we will connect `EditActivity` with everything else.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuestionFragment.java
```diff
@@ -0,0 +1,110 @@
...
+
+        // We have to change the question title and text
+        String questionTitleStr = "Question " + (index + 1) + " - " + question.getTitle();
+        editQuestionDisplay.setText(question.getText());
+        editQuestionTitle.setText(questionTitleStr);
+
+        prepareEditAnswerFormatFragment(question, index);
+    }
+
+    private AnswerModel getNewAnswerModel(AnswerFormat answerFormat, int index) {
```
nbaehler, line 78: The model is not always new so maybe the methods name is a bit misleading.

app/src/androidTest/java/ch/epfl/qedit/edit/EditOverviewFragmentTest.java
```diff
@@ -22,6 +26,21 @@
     public final FragmentTestRule<?, EditOverviewFragment> testRule =
             FragmentTestRule.create(EditOverviewFragment.class);
 
+    @Rule
```
nbaehler, line 29: How come that you added these lines but the tests themselves didn't change? Perhaps I'm missing something here.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuestionFragment.java
```diff
@@ -0,0 +1,110 @@
...
+                        getViewLifecycleOwner(),
+                        new Observer<Integer>() {
+                            @Override
+                            public void onChanged(Integer index) {
+                                editOnQuestionChanged(model.getQuiz(), index);
+                            }
+                        });
+    }
+
+    private boolean testOnQuiz(Quiz quiz, Integer index) {
```
nbaehler, line 58: I would maybe use another name for that method like `quizAndIndexAreValid()` but that's only a suggestion.

app/src/androidTest/java/ch/epfl/qedit/edit/EditOverviewFragmentTest.java
nbaehler, line 29: How come that you had to add these lines when the tests themselves did not change? Maybe I'm missing something here.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuestionFragment.java
nbaehler, line 58: I would maybe use another name for this method, like `quizAndIndexAreValid` but that is only a suggestion.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuestionFragment.java
nbaehler, line 78: And here, not always a new Model is created so the name is maybe a bit misleading.

app/src/main/java/ch/epfl/qedit/view/edit/EditQuizActivity.java
```diff
@@ -2,19 +2,49 @@
...
 public class EditQuizActivity extends AppCompatActivity {
+    private QuizViewModel model;
+    private Quiz quiz;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_quiz);
 
+        tempQuiz();
```
nbaehler, line 23: There is a method in the androidTests under Util that return a standard quiz for tests etc. It could be of use for you in the future when need a quiz for testing. By the way didn't you prepared the code that get the quiz from a bundle? I think that this will be the way how we pass around the quiz, like in the QuizActivity.

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
```diff
@@ -60,17 +65,39 @@ public String getText(Question item) {
...
+                            default:
+                                break;
+                        }
                     }
                 });
     }
 
+    private void setFragment(Fragment fragment) {
+        getActivity()
```
nbaehler, line 94: `RequireActivity()` would be safer I think.

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
nbaehler: Yes I think there is indeed a problem here.

app/src/androidTest/java/ch/epfl/qedit/edit/EditOverviewFragmentTest.java
cjordan7: That's because EditOverviewFragment depends on the EditActivity itself. Each time it gave me a null pointer exception. This resolved it.

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
cjordan7: You were right. I resolved the problem.

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
cjordan7: True

app/src/main/java/ch/epfl/qedit/view/edit/EditQuizActivity.java
cjordan7: Oh yes, but since nothing is connected with everything else. I created some dummy one. This should be deleted in the future and replaced.

PR #92, opened 2020-04-05
Consider refactoring the Visitor into an abstract class

brunnerant:
This PR tackles issue #86 and removes the visitor from the answer format, because it is "too overkill". Instead, we will be using abstract methods, which does the same thing in our situation, but in a more concise manner.


app/src/main/java/ch/epfl/qedit/model/answer/MultiFieldFormat.java
```diff
@@ -21,6 +22,16 @@ public MultiFieldFormat(AnswerFormat... formats) {
...
     }
 
+    @Override
+    public AnswerModel getEmptyAnswerModel() {
+        return null;
+    }
+
+    @Override
+    public AnswerFragment getAnswerFragment() {
+        return null;
```
nathan96g, line 32: we return null because it's unnimplemented right ?

PR #95, opened 2020-04-11
Improve the firestore database interface to load quiz

Anthony-Iozzia:


app/src/test/java/ch/epfl/qedit/backend/MockDBTest.java
```diff
@@ -37,6 +70,125 @@ private void lockWait() {
...
+                new Callback<Response<List<String>>>() {
+                    @Override
+                    public void onReceive(Response<List<String>> data) {
+                        languages = data.getData();
+                        error = data.getError();
+                    }
+                });
+        lockWait();
+        assertEquals(Response.NO_ERROR, error);
+        assertTrue(languages.containsAll(Arrays.asList("en", "fr")));
```
brunnerant, line 153: I think that Harmcrest's `assertThat(languages, containsInAnyOrder(Arrays.asList("en", "fr")));` checks that the list contains exactly the two languages, so you could merge your two lines.

app/src/main/java/ch/epfl/qedit/backend/database/DatabaseService.java
```diff
@@ -18,6 +19,47 @@
...
+    /**
+     * Asynchronously retrieves a supported language table of a quiz from the database. Note that we
+     * assume for convenience that that database model is key-based, because it is the case for
+     * Firestore.
+     *
+     * @param quizID the id of the quiz in the database from which we retrieve the supported
+     *     language table
+     * @param responseCallback the callback that will be triggered when the data arrives
+     */
+    void getSupportedLanguage(
```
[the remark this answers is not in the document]
nathan96g: yes true, orthographic error

app/src/test/java/ch/epfl/qedit/backend/MockDBTest.java
nathan96g: Yes i know, i have tried this but i containsInAnyOrder is not recognize and i don't know why

brunnerant:
Note that I disabled the status checks to merge this PR. This is because it was a backend-only PR, and it was impossible to raise the diff-coverage above 65% (which is already quite good !), since we cannot test Firestore.

PR #96, opened 2020-04-12
Improve Matrix fragments: add support for already filled fields, maybe parentheses

brunnerant:


brunnerant:
Comment for the reviewers: If anyone has an idea how to reduce the cognitive complexity of `fillTable`and `numericField`, it is welcome !

PR #105, opened 2020-04-30
HomeActivity - Online Activity

cjordan7:


app/src/main/java/ch/epfl/qedit/view/Online/OnlineFragment.java
```diff
@@ -0,0 +1,201 @@
...
+                });
+        super.onCreateOptionsMenu(menu, inflater);
+    }
+
+    // This method will be called when an item of the top bar is clicked on
+    @Override
+    public boolean onOptionsItemSelected(@NonNull MenuItem item) {
+        switch (item.getItemId()) {
+            case android.R.id.home:
+                requireActivity().onBackPressed();
```
brunnerant, line 142: There is a bug: when navigating back to the home activity, the hamburger menu item isn't selected. I think that we need to do something like `navigationView.setCheckedItem(id);`.

app/src/main/java/ch/epfl/qedit/view/Online/OnlineFragment.java
```diff
@@ -0,0 +1,201 @@
...
+        String appLanguage = LocaleHelper.getLanguage(requireContext());
+
+        // If the quiz was translated in the application language, pick that version,
+        // otherwise we pick the first one. Note that we could have some more complex logic
+        // here, for example trying english first, and then falling back.
+        if (languages.contains(appLanguage)) return appLanguage;
+        else return languages.get(0);
+    }
+
+    private Intent save(Bundle b, Quiz q) {
```
brunnerant, line 188: There is a small "bug": when we rotate the screen, the search string isn't saved, so we have to retype it. It would be nice to find a way to save it on a rotation change.

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
```diff
@@ -11,21 +11,22 @@
...
 import ch.epfl.qedit.model.Question;
 import ch.epfl.qedit.model.StringPool;
 import ch.epfl.qedit.view.util.ListEditView;
 import ch.epfl.qedit.viewmodel.EditionViewModel;
 import java.util.ArrayList;
-import java.util.List;
 
 /** This fragment is used to view and edit the list of questions of a quiz. */
 public class EditOverviewFragment extends Fragment {
+    private StringSearchable titles = new StringSearchable();;
```
nbaehler, line 23: Typo

app/src/main/java/ch/epfl/qedit/view/edit/EditQuestionActivity.java
```diff
@@ -159,6 +159,8 @@ private void returnResult() {
             // return the Question created by this activity to the callee activity
             Question question = new Question(titleId, textId, answerFormat);
             Intent intent = new Intent();
+
+            // TODO This below doesn't work and I don't understand why
```
nbaehler, line 163: What is not working here? I'll can fix it as part of my task.

app/src/main/java/ch/epfl/qedit/view/util/ListEditView.java
```diff
@@ -144,9 +142,12 @@ public void onClick(View v) {
      *
      * @param <T> the type of the underlying items
      */
-    public static class Adapter<T> extends RecyclerView.Adapter<ItemHolder> {
+    public static class Adapter<T, E extends SearchablePair<T>>
+            extends RecyclerView.Adapter<ItemHolder> implements Filterable {
 
-        private final List<T> items;
+        private final E e;
```
brunnerant, line 148: I think that if we want to keep the `ListEditView` generic and reusable, it is a bad idea to add fields that only serve the purpose of the online activity. I am not saying you shouldn't do it, but really think about another way before doing that. Since other activities use the `ListEditView`, introducing changes might break things in another place...

brunnerant: What I would recommend is that you subclass `ListEditView` to add the functionality that you need, so that it doesn't have an impact on the rest of the app.

app/src/main/res/layout/activity_online2.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
```
brunnerant, line 1: Is there a particular reason why this file is called `activity_online2` ? If not, I think we can safely remove the 2.

app/src/main/res/layout/activity_online2.xml
```diff
@@ -0,0 +1,47 @@
...
+    android:id="@+id/drawer_layout"
+    android:weightSum="5"
+    tools:context=".view.Online.OnlineActivity">
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="vertical">
+
+        <androidx.fragment.app.FragmentContainerView
+            android:id="@+id/home_info_container"
```
brunnerant, line 16: There is a small style issue here: the home fragment is not centered, but rather justified to the left, which is visually annoying. I think we should try to center it.

brunnerant:
I think that we cannot merge this PR until it has more coverage. Most of the code is about UI, so I really think that it is testable...

app/src/main/java/ch/epfl/qedit/view/edit/EditOverviewFragment.java
```diff
@@ -11,21 +11,22 @@
...
     public static final String QUESTION = "ch.epfl.qedit.view.edit.QUESTION";
     public static final int EDIT_QUESTION_ACTIVITY_REQUEST_CODE = 0;
-
-    private ListEditView.Adapter<String> adapter;
+    private ListEditView.Adapter<String, StringSearchable> adapter;
```
nbaehler, line 26: Is it the goal to be able to search in the edition workflow? Otherwise I do not understand why this part should be affected by the search code.

app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
```diff
@@ -41,14 +43,15 @@
...
+
+    private ListEditView.Adapter<Map.Entry<String, String>, SearchableMapEntry> listAdapter;
 
     private ConfirmDialog deleteDialog;
     private EditTextDialog addDialog;
     private int deleteIndex;
 
     private User user;
-    private List<Map.Entry<String, String>> quizzes;
+    private SearchableMapEntry quizzes = new SearchableMapEntry();
```
nbaehler, line 54: This is a list right? So there are several searchable quizzes but the `SearchableMapEntry` suggests only one entry which is confusing to me.

app/src/main/java/ch/epfl/qedit/backend/database/FirebaseDBService.java
```diff
@@ -70,4 +74,42 @@ private static void error(CompletableFuture<?> future, String message) {
 
         return future;
     }
+
+    public void tryy(
```
nbaehler, line 78: I think this method can be private. And I'm not a big fan of its name ;)

app/src/main/java/ch/epfl/qedit/view/home/HomeActivity.java
```diff
@@ -15,11 +15,13 @@
 import ch.epfl.qedit.R;
 import ch.epfl.qedit.model.User;
 import ch.epfl.qedit.util.LocaleHelper;
+import ch.epfl.qedit.view.Online.OnlineActivity;
 import com.google.android.material.navigation.NavigationView;
 import java.util.Objects;
 
 public class HomeActivity extends AppCompatActivity
         implements NavigationView.OnNavigationItemSelectedListener {
+    // private ListEditView.Adapter<Map.Entry<String, String>, SearchableMapEntry> listAdapter;
```
nbaehler, line 24: Can be removed

app/src/main/java/ch/epfl/qedit/view/Online/OnlineActivity.java
```diff
@@ -0,0 +1,51 @@
...
+public class OnlineActivity extends AppCompatActivity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_online2);
+
+        User user =
+                new User(
+                        ", you are Online. You can Search the database",
```
nbaehler, line 22: There are two commas shown to the user.

app/src/main/java/ch/epfl/qedit/view/Online/OnlineFragment.java
```diff
@@ -0,0 +1,201 @@
...
+        return true;
+    }
+
+    private CompletableFuture<StringPool> stringPool(String string) {
+        return db.getQuizLanguages(string)
+                .thenCompose(languages -> db.getQuizStringPool(string, getBestLanguage(languages)));
+    }
+
+    // Handles when a user clicked on the button to edit a quiz
+    private void editQuiz(int position) {
```
nbaehler, line 155: Should we be able to modify the quizzes from this activity? I do not think that it makes a lot on sense as this is the place to search for quizzes made by others and that are publicly available. Same goes for the removal by the way.

app/src/main/java/ch/epfl/qedit/view/Online/OnlineActivity.java
```diff
@@ -0,0 +1,51 @@
...
+                        "",
+                        User.Role.Administrator);
+
+        // Prepare a bundle that contains the user and create a new HomeInfoFragment
+        Bundle bundle = new Bundle();
+        bundle.putSerializable(USER, user);
+        HomeInfoFragment homeInfoFragment = new HomeInfoFragment();
+        homeInfoFragment.setArguments(bundle);
+
+        // Create a new HomeQuizListFragment with the user in its Arguments
```
nbaehler, line 32: Does not match the code.

brunnerant:
We had some issues with this PR, in particular with searching quizzes in the database. It would require more investigation on how to efficiently search for quizzes using Firestore. For now we will close it because we have stopped working on it. We might reopen it later if we have time.

PR #119, opened 2020-05-04
Extend first part of the edition workflow and add link it to the rest of the app

nbaehler:
Add an activity that allows to do some settings before editing an existing quiz and link the entire edition workflow to the rest of the app. When clicking on plus for for adding a question one will be sent directly to the QuestionActivity, this restricts the flow of edition but it is much clearer and is easier to handle for us. I'll remove the logic that handled having empty questions in the list that aren't edited yet.


app/src/main/java/ch/epfl/qedit/model/StringPool.java
```diff
@@ -12,17 +12,17 @@
...
     private Map<String, String> stringPool;
+    private String languageCode;
 
     public StringPool() {
         stringPool = new HashMap<>();
     }
 
     public StringPool(Map<String, String> stringPool) {
         this.stringPool = new HashMap<>(stringPool);
+        this.languageCode = languageCode;
```
cjordan7, line 25: I don't understand. I think it is equivalent to:

`this.languageCode = this.languageCode;`

app/src/main/java/ch/epfl/qedit/model/StringPool.java
```diff
@@ -72,4 +72,12 @@ public String get(String id) {
 
         return stringPool.get(id);
     }
+
+    public String getLanguageCode() {
+        return languageCode;
+    }
+
+    public void setLanguageCode(String languageCode) {
+        this.languageCode = languageCode;
```
cjordan7, line 81: This I  understand.

app/src/main/java/ch/epfl/qedit/model/StringPool.java
nbaehler: Oh, that's an artefact that I should have removed. I'm going to deleted the line right away.

app/src/main/java/ch/epfl/qedit/view/edit/EditPreviewFragment.java
```diff
@@ -37,15 +36,7 @@ public View onCreateView(
...
-        model.getFocusedQuestion()
-                .observe(
-                        getViewLifecycleOwner(),
-                        new Observer<Integer>() {
-                            @Override
-                            public void onChanged(Integer index) {
-                                onQuestionChanged(index);
-                            }
-                        });
+        model.getFocusedQuestion().observe(getViewLifecycleOwner(), this::onQuestionChanged);
```
nathan96g, line 39: this is cool, i didn't knew this syntax

app/src/main/java/ch/epfl/qedit/view/edit/EditSettingsActivity.java
```diff
@@ -0,0 +1,117 @@
...
+    public void onNothingSelected(AdapterView parent) {
+        // Not used because there will always be something selected
+    }
+
+    @Override
+    /* This method is needed to apply the desired language at the activity startup */
+    protected void attachBaseContext(Context base) {
+        super.attachBaseContext(LocaleHelper.onAttach(base));
+    }
+}
```
nathan96g, line 117: from what i understand, this activity will become a dialog in the next sprint. Right?

app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
```diff
@@ -162,10 +153,43 @@ private void deleteConfirmation(int position) {
...
+    //                            else
+    //                                launchQuizActivity(
+    //                                        quizStructure
+    //                                                .join()
+    //                                                .instantiateLanguage(stringPool.join()));
+    //                        });
+    //
+    //        progressBar.setVisibility(INVISIBLE);
+    //    }
+
```
nathan96g, line 188: Why is all of this commented?

app/src/main/java/ch/epfl/qedit/view/edit/EditPreviewFragment.java
nbaehler: Yeah it's nice! The operator is called 'method reference operator'.

app/src/main/java/ch/epfl/qedit/view/edit/EditSettingsActivity.java
nbaehler: Yes, it will simplify the workflow and will also look nicer. I didn't have enough time to do these changes in this sprint. Additionally we will be able to use `startActivityForResult` for the entire edition by eliminating the settings activity.

app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
nbaehler: This is just temporary, the time constraints did not allow me to integrate this into the workflow. So I couldn't test it and in order to not reduce the coverage unnecessarily I commented this part.

PR #121, opened 2020-05-07
Issue 112, answers correction

alexisc99:


app/src/main/java/ch/epfl/qedit/view/quiz/QuizActivity.java
```diff
@@ -80,6 +86,34 @@ public boolean onOptionsItemSelected(@NonNull MenuItem item) {
             case android.R.id.home:
                 onBackPressed();
                 break;
+            case R.id.validate:
+                new AlertDialog.Builder(this)
```
brunnerant, line 90: I think that we have already created a `ConfirmDialog` in the package `view.util`. If you reuse it, it would probably make that big piece of code much shorter ;) .

app/src/main/java/ch/epfl/qedit/model/answer/MatrixModel.java
```diff
@@ -29,4 +35,22 @@ public String getAnswer(int row, int col) throws IndexOutOfBoundsException {
             throw new IndexOutOfBoundsException();
         }
     }
+
+    @Override
+    public boolean equals(@Nullable Object o) {
+        if (o instanceof MatrixModel) {
```
brunnerant, line 41: If you want to reduce the indentation hell, we can change it to `if (!(o instanceof MatrixModel)) return false;`. For the rest, though, I have no idea how we could reduce the nesting...

app/src/main/java/ch/epfl/qedit/model/answer/EmptyAnswerModel.java
```diff
@@ -1,4 +1,11 @@
 package ch.epfl.qedit.model.answer;
 
+import androidx.annotation.Nullable;
+
 /** This model contains nothing */
-public class EmptyAnswerModel extends AnswerModel {}
```
nbaehler, line 4: With my PR merged you won't need to implement the method as I removed EmptyAnswerFormat and everything that goes with it.

app/src/main/java/ch/epfl/qedit/model/answer/AnswerFormat.java
```diff
@@ -19,6 +19,17 @@
      */
     private String text;
 
+    /** The correct answer */
+    private AnswerModel questionAnswer;
```
nbaehler, line 23: I was a bit confused about the naming. Probably I would have picked `solution` but this is only a suggestion.

app/src/main/java/ch/epfl/qedit/model/answer/MatrixModel.java
nbaehler: I do not know either, maybe you could define a method for the inner most `if`but that's kind of artificial.

app/src/main/java/ch/epfl/qedit/model/answer/AnswerFormat.java
alexisc99: ahaha Of course !

app/src/main/java/ch/epfl/qedit/model/answer/EmptyAnswerModel.java
alexisc99: Okok I will delete this too, thanks

app/src/main/java/ch/epfl/qedit/model/answer/MatrixModel.java
alexisc99: I will try something with a boolean that is true at first and then an "&=" operator for each loop iteration.

app/src/main/java/ch/epfl/qedit/view/quiz/QuizActivity.java
alexisc99: Yes, thank you !

app/src/androidTest/java/ch/epfl/qedit/quiz/QuizActivityTest.java
```diff
@@ -138,6 +138,37 @@ public void testTimeIsClicked() {
         finishActivity();
     }
 
+    @Test
```
brunnerant, line 141: I think that we should add some actual checks to the tests. For example, check that the confirmation dialog is shown when clicking on the validate button.

app/src/androidTest/java/ch/epfl/qedit/quiz/QuizActivityTest.java
```diff
@@ -138,6 +138,37 @@ public void testTimeIsClicked() {
...
+        launchActivity();
+        onView(withId(R.id.validate)).perform(click());
+        finishActivity();
+    }
+
+    @Test
+    public void testDoneNoCLicked() {
+        launchActivity();
+        onView(withId(R.id.validate)).perform(click());
+        // onView(withText("No")).inRoot(isDialog()).perform(click());
```
nbaehler, line 152: Are these comments intended?

app/src/androidTest/java/ch/epfl/qedit/quiz/QuizActivityTest.java
brunnerant: Ok I just realized this is covered by another test of the same class, so I guess it is ok.

app/src/androidTest/java/ch/epfl/qedit/quiz/QuizActivityTest.java
alexisc99: They can be safely removed

Issue #129, opened 2020-05-10
Add a backend interface for the location service, and mock it


brunnerant:
This task will be useful for the next tasks that will use the location. Once the location service is behind an interface, we will be able to test it by creating a mock.

PR #130, opened 2020-05-10
Add a backend interface for the location service, and mock it

brunnerant:
This PR is one of the tasks towards user story #123. It adds a location service interface to the codebase, so that it can easily be mocked. Closes issue #129.


app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
```diff
@@ -0,0 +1,34 @@
+package ch.epfl.qedit.util;
+
+import android.content.Context;
```
cjordan7, line 3: I don't think it is a good idea to put that class in `util`.

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class LocationHelper {
+    public static void initLocationProvider(Context context) {
+        LocationManager manager =
+                (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+
+        if (!manager.isProviderEnabled(MockLocService.LOCATION_PROVIDER)) {
+            manager.addTestProvider(
+                    MockLocService.LOCATION_PROVIDER,
+                    false,
```
cjordan7, line 16: I think it would be great to explain what this army of `falses` is.

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
```diff
@@ -0,0 +1,34 @@
+package ch.epfl.qedit.util;
+
+import android.content.Context;
+import android.location.Criteria;
+import android.location.LocationManager;
+import ch.epfl.qedit.backend.location.MockLocService;
+
+public class LocationHelper {
+    public static void initLocationProvider(Context context) {
```
cjordan7, line 9: I don't understand what this class is for. Please add more comments !

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
```diff
@@ -0,0 +1,34 @@
...
+                    false,
+                    false,
+                    Criteria.POWER_LOW,
+                    Criteria.ACCURACY_FINE);
+            manager.setTestProviderEnabled(MockLocService.LOCATION_PROVIDER, true);
+        }
+    }
+
+    public static void cleanupLocationProvider(Context context) {
+        LocationManager manager =
```
cjordan7, line 30: I'm not even sure what those function even do... add comments !

app/src/main/java/ch/epfl/qedit/backend/location/AndroidLocService.java
```diff
@@ -0,0 +1,47 @@
...
+import android.Manifest;
+import android.content.Context;
+import android.content.pm.PackageManager;
+import android.location.Criteria;
+import android.location.LocationListener;
+import android.location.LocationManager;
+import androidx.core.app.ActivityCompat;
+
+public class AndroidLocService implements LocationService {
+
```
cjordan7, line 12: Why call it `AndroidLocService` and not `LocService` ?

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
```diff
@@ -0,0 +1,61 @@
...
+    /** This is the id of the mock location provider */
+    public static final String LOCATION_PROVIDER = "mock_provider";
+
+    // This is the set of listeners that need to be updated.
+    private final Set<LocationListener> listeners;
+
+    // This indicates whether the user has permissions to access this service.
+    // Note that this is for test purposes. It allows to test how the UI reacts when the user
+    // doesn't have the access to the location.
+    private boolean permission;
```
cjordan7, line 20: It has a bad name for a boolean.

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
```diff
@@ -0,0 +1,61 @@
...
+        Location location = new Location(LOCATION_PROVIDER);
+        location.setLongitude(longitude);
+        location.setLatitude(latitude);
+
+        // And then we trigger the listeners
+        for (LocationListener listener : listeners) listener.onLocationChanged(location);
+    }
+
+    /** Sets the permission to access the location service. */
+    public void setPermission(boolean permission) {
```
cjordan7, line 58: Why don't you give full comments (params,...) for this function, sometimes you do and sometimes you don't... There is a lack of consistency in this.

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
```diff
@@ -0,0 +1,61 @@
...
+    private boolean permission;
+
+    public MockLocService(Context context) {
+        listeners = new HashSet<>();
+        permission = false;
+    }
+
+    @Override
+    public boolean subscribe(LocationListener listener, int interval) {
+        // In the mock, we will not take into account the interval, because the goal
```
cjordan7, line 29: You don't use `interval` ?

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
```diff
@@ -0,0 +1,61 @@
...
+        // is just to test how the UI responds to the location updates.
+        if (permission) listeners.add(listener);
+
+        return permission;
+    }
```
cjordan7, line 34: So you never change the `permission` ...

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public MockLocService(Context context) {
+        listeners = new HashSet<>();
+        permission = false;
+    }
+
+    @Override
+    public boolean subscribe(LocationListener listener, int interval) {
+        // In the mock, we will not take into account the interval, because the goal
+        // is just to test how the UI responds to the location updates.
```
cjordan7, line 30: The more I look at it and the more I think that function is odd. First `interval` is not used. Second if you want to subscribe. You have to give permission. But then if someone else wants to subscribe and you don't want so you have to call `setPermission`

So it would be something like
`setPermission(True)`
`subscribe(...)`
`setPermission(False)`
`subscribe(...)`

Oh wait... why would you want to subscribe someone who you know wouldn't have permission.

app/src/test/java/ch/epfl/qedit/ExampleUnitTest.java
```diff
@@ -1,17 +0,0 @@
...
-
-/**
- * Example local unit test, which will execute on the development machine (host).
- *
- * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
- */
-public class ExampleUnitTest {
-    @Test
-    public void addition_isCorrect() {
-        assertEquals(4, 2 + 2);
```
cjordan7, line 15: I understand why you deleted this class

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
```diff
@@ -0,0 +1,26 @@
...
+     * @return true if the permissions were granted for the location service, false otherwise.
+     */
+    boolean subscribe(LocationListener listener, int interval);
+
+    /**
+     * Unsubscribe the given listener, meaning that it will stop receiving updates.
+     *
+     * @param listener the location listener to unsubscribe.
+     */
+    void unsubscribe(LocationListener listener);
```
cjordan7, line 25: Why not returning a boolean as well. I think this is inconsistent with `subscribe`

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
```diff
@@ -0,0 +1,26 @@
...
+ * This interface is used to request location updates from the phone. It is an interface, so that it
+ * can be mocked for the tests.
+ */
+public interface LocationService {
+    /**
+     * Subscribes to the location service using the given location listener. The location updates
+     * should be done at the given interval (in milliseconds), if possible.
+     *
+     * @param listener the listener that will be subscribed.
+     * @param interval the interval at which updates should be received.
```
cjordan7, line 15: I don't quite understand this parameter (`interval`)... Furthermore, the function is called `subscribe` and you want to send updates but where is the function for updates ?

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
```diff
@@ -0,0 +1,26 @@
...
+     * @return true if the permissions were granted for the location service, false otherwise.
+     */
+    boolean subscribe(LocationListener listener, int interval);
+
+    /**
+     * Unsubscribe the given listener, meaning that it will stop receiving updates.
+     *
+     * @param listener the location listener to unsubscribe.
```
cjordan7, line 23: This sentence feels odd `the location listener to unsubscribe`

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
```diff
@@ -0,0 +1,26 @@
+package ch.epfl.qedit.backend.location;
+
+import android.location.LocationListener;
+
+/**
+ * This interface is used to request location updates from the phone. It is an interface, so that it
+ * can be mocked for the tests.
+ */
+public interface LocationService {
+    /**
```
cjordan7, line 10: Why is it in `backend` package ?

app/src/androidTest/java/ch/epfl/qedit/backend/MockLocTest.java
```diff
@@ -0,0 +1,66 @@
...
+                    public void onStatusChanged(String provider, int status, Bundle extras) {}
+
+                    @Override
+                    public void onProviderEnabled(String provider) {}
+
+                    @Override
+                    public void onProviderDisabled(String provider) {}
+                };
+
+        locService.subscribe(listener, 42);
```
cjordan7, line 61: You don't test what `subscribe` return ?

app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,10 @@
     package="ch.epfl.qedit">
 
     <uses-permission android:name="android.permission.CAMERA"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
```
cjordan7, line 8: What are `ACCESS_FINE_LOCATION`, `ACCESS_COARSE_LOCATION`, `ACCESS_NETWORK_STATE`. What are they used for ? Please add some comments.

app/src/main/java/ch/epfl/qedit/backend/location/AndroidLocService.java
brunnerant: Because I have the Mock service, and the real service... I wanted to make the difference clear, but that might be some bad naming on my part. I have to think about another name ;-).

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
brunnerant: Yeah, I just put full comments when the function is not obvious. In this case I think that it is pretty clear what it does.

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
brunnerant: No, but I added it for the real location service, so that the location is not updated too quickly. But in the mock, I want to be able to set the location wherever I want.

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
brunnerant: The function `setPermission` is responsible for changing the permission.

app/src/main/java/ch/epfl/qedit/backend/location/MockLocService.java
brunnerant: Ok, let me clarify what permission is. In the real location service, we will need to request the user for the permission to use the location services. I modelled that by returning a boolean if the operation was allowed. That way, in the activities that use the location service, I can request the permission if I see that this method returned false.
Now, in the mock version, I had to artificially model permissions using a boolean flag. That way, I will be able to see how the app reacts if I disable permissions for example.

app/src/test/java/ch/epfl/qedit/ExampleUnitTest.java
brunnerant: Yeah, I didn't notice it was there until now, so I thought that it was rather safe deleting it ;).

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
brunnerant: Mhhh because unsubscribing is "safe" in the sense that you cannot unsubscribe from the location service if you didn't subscribe to it in the first place.

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
brunnerant: The listener is the one to receive the updates. The `interval` determines the interval at which we would ideally like to receive the updates.

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
brunnerant: Because it is not part of the UI, nor the model. It is an interface that talks with services external to the application, so it feels pretty natural to put it there.

app/src/androidTest/java/ch/epfl/qedit/backend/MockLocTest.java
brunnerant: That's a good point, I will add that.

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
brunnerant: I didn't find a better place for now... If you have a suggestion, tell me ;).

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
brunnerant: Yeah, you're right. In android studio, you can see the name of the arguments, which makes it clearer, but not on GitHub, so I will add it.

app/src/androidTest/java/ch/epfl/qedit/util/LocationHelper.java
brunnerant: Okay, you are right.

app/src/main/java/ch/epfl/qedit/backend/location/LocationService.java
brunnerant: Yes it feels odd, but it carries the meaning that I want ;).

brunnerant:
Yes it would be nice to have more coverage, but I don't think it is possible. I already have 100% coverage on the mock location service, so I cannot increase the diff coverage anywhere else ;-(.

.github/workflows/android.yml
```diff
@@ -48,7 +48,10 @@ jobs:
           api-level: 24
           arch: x86_64
           disable-animations: true
-          script: ./gradlew build connectedCheck jacocoTestReport
+          script: |
```
nbaehler, line 51: I have been following your discussion thread on the forum. Am I right that there is still one problem with the Mock?

brunnerant: No, I managed to solve it, and I posted those lines of code if anyone needs it. What it does is that it first installs the app on the emulator, then grants permission for mocking the location, and finally launches the tests.

PR #134, opened 2020-05-11
Add choice for TreasureHunt in EditSettings

cjordan7:


cjordan7:
> It is a good first step towards location in treasure hunt quizzes. There is just one thing that I am wondering about. Treasure hunt quizzes should contain a location for every question. This means that the user shouldn't have the choice to check or uncheck the location. Will you handle that in a future PR, or do you plan on doing it now ?

Yeah, I think I will change it. I will rename location to TreasureHunt checkbox and the user will have to give a location

app/src/main/java/ch/epfl/qedit/model/Quiz.java
```diff
@@ -65,6 +65,11 @@ public Builder(Quiz quiz) {
             this.treasureHunt = quiz.treasureHunt;
         }
 
+        public Builder setTreasureHunt(Boolean isTreasureHunt) {
```
brunnerant, line 68: I don't think it is a really good idea to have that method here. That means that you can change the treasure hunt boolean in the middle of building a quiz, meaning that you could make the questions invalid.

app/src/main/res/layout/activity_edit_modify_quiz_settings.xml
```diff
@@ -21,6 +21,19 @@
         app:layout_constraintBottom_toTopOf="@+id/button_start_editing"
         android:inputType="text" />
 
+    <CheckBox
+        android:id="@+id/treasure_hunt_checkbox"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/regular_padding"
+        android:text="@string/treasureHunt"
```
brunnerant, line 29: I think that the convention is to use `snake_case` in the xml files.

app/src/main/res/layout/activity_edit_new_quiz_settings.xml
```diff
@@ -21,6 +21,19 @@
         app:layout_constraintBottom_toTopOf="@+id/edit_language_selection"
         android:inputType="text" />
 
+    <CheckBox
+        android:id="@+id/treasure_hunt_checkbox"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/regular_padding"
+        android:text="@string/treasureHunt"
```
brunnerant, line 29: Ssssnake cassssse.

app/src/main/res/layout/activity_edit_modify_quiz_settings.xml
```diff
@@ -21,6 +21,19 @@
         app:layout_constraintBottom_toTopOf="@+id/button_start_editing"
         android:inputType="text" />
 
+    <CheckBox
```
brunnerant, line 24: How do you handle the case when a non-treasure hunt is converted to a treasure hunt quiz ? It means that the user must add a location to all the questions. I think that the easiest thing for now is to not allow the user to change that setting.

PR #137, opened 2020-05-12
Modify User class by adding statistic attributes

nathan96g:
- add success, attempt and score attribute to user class
- remove Role attribute


app/src/main/java/ch/epfl/qedit/model/User.java
```diff
@@ -86,17 +94,43 @@ public String getFullName() {
...
+    public int getSuccess() {
+        return success;
+    }
+
+    public int getAttempt() {
+        return attempt;
+    }
+
+    public void incrementScore(int points) {
+        if (points < 0)
```
cjordan7, line 110: Maybe we could have penality points one day

app/src/main/java/ch/epfl/qedit/model/User.java
```diff
@@ -86,17 +94,43 @@ public String getFullName() {
...
+            throw new IllegalArgumentException("Cannot increment score with negative points");
+        score += points;
```
cjordan7, line 112: Maybe when we increment score we should increment success and attempt in the same time

app/src/main/java/ch/epfl/qedit/model/User.java
nathan96g: Yes okey, i will make the condition (score - points) >= 0

app/src/main/java/ch/epfl/qedit/model/User.java
nathan96g: to be fair, this i am not sure yet, i prefer to let the personne who use the user class to decide how he want to handle the increment of score. However i can increment attempte when i increment success

app/src/main/java/ch/epfl/qedit/model/User.java
brunnerant: (In this case the condition should be `score + points >= 0`)

app/src/main/java/ch/epfl/qedit/model/User.java
```diff
@@ -24,13 +19,26 @@
 
     private final String firstName;
     private final String lastName;
-    private final Role role;
 
-    public User(String firstName, String lastName, Role role) {
+    private int score;
+    private int success;
```
brunnerant, line 24: I feel like we could find more explicit names instead of `success` and `attempt`. Like `numSuccesses` and `numAttempts`, or `successes` and `attempts`...

app/src/main/java/ch/epfl/qedit/model/User.java
nathan96g: yes yes it s just that to keep nice method names i have added a decrementScore and here the condition is score - point

app/src/main/java/ch/epfl/qedit/model/User.java
brunnerant: Otherwise, LGTM. It was nice to remove the roles, since we don't use them anymore !

brunnerant:
LGTM.

app/src/main/java/ch/epfl/qedit/model/User.java
nathan96g: Yes right, my bad, it's my english ^^

app/src/main/java/ch/epfl/qedit/model/User.java
brunnerant: No problem ;).

app/src/main/java/ch/epfl/qedit/model/User.java
brunnerant: No problem ;).

PR #141, opened 2020-05-12
Test ScannerActivity and QRGeneratorFragment as much as possible

alexisc99:
…dencies


app/src/main/java/ch/epfl/qedit/view/QR/ScannerActivity.java
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                Toast.makeText(this, "Permission Denied", Toast.LENGTH_LONG).show();
+                super.onBackPressed();
+            }
+        }
+    }
+
+    @Override
+    public void handleResult(Result result) {
+        /*Log.d("QRCodeScanner", result.getText());
```
brunnerant, line 97: Can you make sure to remove the debug comments before merging ? thanks !

app/src/main/java/ch/epfl/qedit/view/QR/QRGeneratorFragment.java
```diff
@@ -0,0 +1,55 @@
...
+        if (quiz_name.length() > 0) {
+            WindowManager manager =
+                    (WindowManager) requireActivity().getSystemService(WINDOW_SERVICE);
+            Display display = manager.getDefaultDisplay();
+            Point point = new Point();
+            display.getSize(point);
+            int width = point.x;
+            int height = point.y;
+            int smallerDimension = width < height ? width : height;
+            smallerDimension *= 3 / 4;
```
brunnerant, line 44: This will be the dimension of the QR code, right ? I am not sure if I understand why we need this.

app/src/main/res/layout/fragment_q_r_generator.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".view.QR.QRGeneratorFragment">
+
+    <ImageView
+        android:layout_width="200dp"
```
brunnerant, line 9: Have you tried the layout on a big screen ? I feel like the fixed size might create a really small QR code on big screens. Or maybe this is needed, I don't know...

app/src/main/res/values/strings.xml
```diff
@@ -128,4 +128,7 @@
     <string name="database_error">An error occurred connecting to the database</string>
     <string name="card_icon">card icon</string>
 
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
brunnerant, line 132: I couldn't find a use of that string... If it is the case, I think we can safely delete it.

alexisc99: I don't know where this come from ahaha. I'll remove it

app/src/main/java/ch/epfl/qedit/view/QR/ScannerActivity.java
alexisc99: Yep

app/build.gradle
```diff
@@ -42,6 +44,7 @@ android {
         }
     }
     testOptions {
+        //execution 'ANDROIDX_TEST_ORCHESTRATOR'
```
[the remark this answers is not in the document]
alexisc99: I kept it as this could help us with testing permissions if android studio 3.6.2 came out before the end of the semester but as the tests pass this way I'll remove it

app/src/main/java/ch/epfl/qedit/view/QR/QRGeneratorFragment.java
alexisc99: It is to make the QR code 3/4 of the size of the smaller length between width and height of the display.

app/src/main/java/ch/epfl/qedit/view/QR/QRGeneratorFragment.java
```diff
@@ -0,0 +1,55 @@
...
+    Bitmap bitmap;
+    QRGEncoder encoder;
+
+    @Override
+    public View onCreateView(
+            LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+        View view = inflater.inflate(R.layout.fragment_q_r_generator, container, false);
+
+        qr_code = (ImageView) view.findViewById(R.id.qr_code);
+        quiz_name = requireArguments().getString(QUIZ_NAME);
```
nathan96g, line 34: The QR code must be generated from a quizID and not a quizName, i understand that the implementation is still the same because it is still a string, but could you just replace QUIZ_NAME by QUIZ_ID? in order to avoid confusion

app/src/main/java/ch/epfl/qedit/view/QR/QRGeneratorFragment.java
```diff
@@ -0,0 +1,55 @@
...
+import android.widget.ImageView;
+import androidmads.library.qrgenearator.QRGContents;
+import androidmads.library.qrgenearator.QRGEncoder;
+import androidx.fragment.app.Fragment;
+import ch.epfl.qedit.R;
+import com.google.zxing.WriterException;
+
+public class QRGeneratorFragment extends Fragment {
+    public static final String QUIZ_NAME = "ch.epfl.qedit.view.quiz_name";
+    String quiz_name;
```
nathan96g, line 23: maybe rename this attributes quiz_id (see comment below)

app/src/main/res/values/strings.xml
nathan96g: it is created automatically when you create a new fragment, deleting is nice :)

app/src/main/res/layout/fragment_q_r_generator.xml
alexisc99: It is large enough, if I make it bigger it might be "flou"

app/src/main/java/ch/epfl/qedit/view/QR/QRGeneratorFragment.java
alexisc99: Yes of course

PR #144, opened 2020-05-12
Panel that display statistics about the current user

nathan96g:
- display a card with QEDit score
- card with number of quizzes succeeded
- card with number of quiz attpemted

Also add here and there in the application the scroll bar in the scrollview to make it more intuitive for the user


app/src/main/res/values/strings.xml
```diff
@@ -55,6 +55,12 @@
     <string name="title_activity_home">Home</string>
     <string name="welcome">Welcome,</string>
 
+    <string name="display">Display</string>
```
brunnerant, line 58: I think that when the card view is hidden, the button should be named "Show statistics". Otherwise, the user cannot guess what the button will open. Same for the french version.

app/src/main/res/layout/fragment_home_info.xml
```diff
@@ -1,17 +1,81 @@
...
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintLeft_toLeftOf="parent"
-        app:layout_constraintRight_toRightOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
+        android:padding="@dimen/regular_padding"
+        android:textStyle="bold"
+        android:layout_margin="@dimen/regular_padding"
+        android:layout_gravity="center"/>
+
+    <TextView
```
brunnerant, line 20: Could you add a little bit of margin under the text view ? I felt like it was a bit to close to the horizontal separator.

app/src/main/res/values/strings.xml
nathan96g: I thought about that and I think that the user can guess easily that it is about statistics because it is display by default when you arrive in the home page. Anyway, you are maybe right, we are never to careful regarding the guessing of a user.

app/src/main/res/layout/fragment_home_info.xml
nathan96g: Of course sire, beautiful padding on its way!

app/src/main/res/values-fr/strings.xml
```diff
@@ -27,6 +27,11 @@
     <string name="title_activity_home">Accueil</string>
     <string name="welcome">Bienvenue,</string>
 
+    <string name="display_stats">Afficher les statistiques</string>
+    <string name="hide_stats">Cacher les statistiques</string>
+    <string name="score">QEDit SCORE</string>
+    <string name="successes">NOMBRE DE QUIZZES RÉUSSIS</string>
```
Anthony-Iozzia, line 33: "QUIZ" :)

Issue #145, opened 2020-05-13
Create an interface for android permissions so that it can be mocked


brunnerant:
Alexis and I need to use android permissions to use the camera and the location. The problem is that it is hard to test the permissions, because they are persistent, and so it is hard to reproduce some of the tests once the emulator was launched. To circumvent this, I decided to hide the details behind an interface that we will mock. That way, we will be able to test arbitrary scenarios.

PR #146, opened 2020-05-13
 Create an interface for android permissions so that it can be mocked

brunnerant:
Please see issue #145 for a detailed explanation of this PR.


app/src/main/java/ch/epfl/qedit/backend/permission/AndroidPermManager.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public boolean shouldAskAgain(PermissionActivity activity, String permission) {
+        return ActivityCompat.shouldShowRequestPermissionRationale(activity, permission);
+    }
+
+    @Override
+    public void requestPermissions(
+            PermissionActivity activity, String[] permissions, OnPermissionResult callback) {
+        callbacks.put(nextRequestCode++, callback);
```
cjordan7, line 39: It is to be sure. But I hope you realize that
`callbacks.put(nextRequestCode++, callback);`

is equivalent to

`callbacks.put(nextRequestCode, callback);`
`nextRequestCode = nextRequestCode + 1;`

brunnerant: Yes, that's what I want to do, but thanks for carefully checking that one ! Maybe I can write it in two lines for more clarity if you want.

PR #147, opened 2020-05-13
Replace SettingsActivity by dialog

nbaehler:
For now I reused some code of `EditTextDialog`. We do not need this class any more but maybe it will be of use later on, so I kept it. Either we'll remove it or I will have to do some refactoring.

But finally I did delete `EditTextDialog` and reused some of the code in the new `Dialog`. Additionally I could remove all the files linked to the `SettingsActivitiy` as we no longer need them.


app/src/main/java/ch/epfl/qedit/view/edit/EditQuizSettingsDialog.java
```diff
@@ -0,0 +1,294 @@
...
+        languageCode = getResources().getStringArray(R.array.languages_codes)[pos];
+    }
+
+    @Override
+    public void onNothingSelected(AdapterView parent) {
+        // Not used because there will always be something selected
+    }
+
+    /**
+     * Gets all the arguments that are passes to the dialog, inflates the view and sets up the
```
brunnerant, line 165: There are a few typos in this comment (I don't even know why I am looking at typos, but anyway).

nbaehler: I'll correct the typos, thanks for point them out!

PR #151, opened 2020-05-14
Remove questions that do not use StringPools, add StringPools where needed and remove the support of the old version of the Questions

nbaehler:
Use only questions that do not store the titles and texts directly any more. Instead add `StringPools` where needed. Additionally I unified the usage of quiz titles across the app. Every user should now have the same quizzes in the `AuthService` and in the `DBService`.

I had to add a `StringPool` to the `QuizViewModel`.

### DIFF
--- a/app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
+++ b/app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
@@ -8,11 +8,11 @@ import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
-import androidx.lifecycle.Observer;
 import androidx.lifecycle.ViewModelProvider;
 import ch.epfl.qedit.R;
 import ch.epfl.qedit.model.Question;
 import ch.epfl.qedit.model.Quiz;
+import ch.epfl.qedit.model.StringPool;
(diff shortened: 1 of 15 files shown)


app/src/main/java/ch/epfl/qedit/view/quiz/QuestionFragment.java
```diff
@@ -44,12 +44,7 @@ public View onCreateView(
                 .getFocusedQuestion()
                 .observe(
                         getViewLifecycleOwner(),
-                        new Observer<Integer>() {
```
brunnerant, line 47: I like when your replace those ugly anonymous classes, thank you !

nbaehler: Oh yes, I missed this one, thx!

nbaehler:
We were a bit too fast on this PR. In the part were we simply answer quizzes we actually do not need to have a `StringPool`. @brunnerant had prepared a function `instantiateLanguage(StringPool pool)` (I had totally forgot it) that replaces the keys by their values in the quiz, questions, ... etc. recursively. Hence I will unwind the changes that I made in this part and make it work with this function. Sorry for the trouble!

nbaehler:
So, I simply took the version on master of the 3 files that handle the answering of quizzes part and added several `instantiateLanguage` in the tests. Works like a charm! Once the CI terminates I will merge this PR.

PR #157, opened 2020-05-15
Add AnswerFragment to EditPreviewFragment

nbaehler:
Finally support the answers with potentially pre-filled fields.


nbaehler:
I somehow branched off from another branch instead of master when creating the branch for the issue. I'm going to reset the branch and push only the last commit.

Issue #165, opened 2020-05-16
Create a treasure hunt activity


brunnerant:
The question locator activity and the question fragment should be put together to create a treasure hunt activity. The user first have to search for the next question, then he can answer it, then search for the next one, and so on...

Issue #128, opened 2020-05-09
Add a fragment to visualize where the next question of a treasure hunt quiz is


brunnerant:
Now that we have added treasure hunt quizzes to the model, we can use them to create an activity to answer them. The goal of this task is to present to the user where the location of the next question is.

PR #138, opened 2020-05-12
 Add a fragment to visualize where the next question of a treasure hunt quiz is

brunnerant:
This PR is one of the tasks towards user story #123. It closes issue #128.
The goal is to create an activity that retrieves the phone's location and tells the user where the next question is relative to his position.
Note that this activity will not yet be linked to the rest of the app, but it will come in future PRs.

PR #166, opened 2020-05-16
Sprint11/issue#165

brunnerant:
See issue #165 for a description of this PR. It depends on issue #128, but I already start working on it, since that issue is at the review stage.


brunnerant:
Ok, I think it is a bad idea to start the pull request before #138 is merged, since it creates a lot of unwanted history. So for now I will close this PR and re-open it later.

PR #167, opened 2020-05-17
Create a treasure hunt activity

brunnerant:
The goal of this PR is to use the question locator activity and link with the already existing question fragment, in order to create a fully usable treasure hunt activity. See #165 for more details.


brunnerant:
I have already written it in the code, but I will repeat it here for clarity. I have added a *temporary* button in the question locator activity that allows to do as if the phone was approaching the question. That's useful for the demos, since it will be a bit hard to move in real life in the middle of a zoom call. It will be easy to remove that later, so it is not a problem at all.

app/src/main/java/ch/epfl/qedit/model/Question.java
```diff
@@ -39,7 +48,10 @@ public Question(String title, String text, AnswerFormat format) {
         this.title = Objects.requireNonNull(title);
         this.text = Objects.requireNonNull(text);
         this.format = Objects.requireNonNull(format);
-        this.location = null;
+
+        // We store some dummy values, since nobody should try to access them anyway
+        this.longitude = 0;
```
nbaehler, line 53: This might not be the cleanest way of doing it but I totally get why we do this. Ideally we might create to types of `Questions`...

app/src/androidTest/java/ch/epfl/qedit/treasurehunt/TreasureHuntActivityTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private void checkIntents(int n) {
+        intended(
+                allOf(
+                        hasComponent(QuestionLocatorActivity.class.getName()),
+                        hasExtra(
+                                equalTo(QuestionLocatorActivity.QUESTION_LONGITUDE), equalTo(42.0)),
+                        hasExtra(equalTo(QuestionLocatorActivity.QUESTION_LATITUDE), equalTo(43.0)),
+                        hasExtra(equalTo(QuestionLocatorActivity.QUESTION_RADIUS), equalTo(100.0))),
+                times(n));
```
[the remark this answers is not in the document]
brunnerant: Yes this line is a bit tricky. In fact, I wanted to do as you said in the first place, but the test framework doesn't work that way unfortunately. In fact when you test the intents using `intended`, it tests all the previous intents, and it doesn't "reset" them. That means that you cannot test that an intent was launched twice in the same test. Instead, you have to test once that two intents where launched. That sucks, but the test framework forced me to do that ;(.

app/src/main/java/ch/epfl/qedit/model/Question.java
```diff
@@ -14,20 +13,30 @@
 
     /**
      * Treasure hunt questions also have a location and a radius within which they can be answered.
+     * android.Location is not serializable, so we store the longitude and latitude instead.
```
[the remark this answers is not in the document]
brunnerant: Yeah, I had to do that move, because I couldn't pass the quiz in the bundle because of that. It sucks that the android API is split between `Parcelable` and `Serializable`. I wish they could have a unified API, but again I was disappointed. They prefer having two APIs and deprecating one, it seems ;).

app/src/main/java/ch/epfl/qedit/model/Question.java
brunnerant: Yeah, I agree with you. Now there is only a boolean in the `Quiz`class to make the difference with treasure hunt quizzes. I could also put a boolean in the `Question` class, but it would be redundant, so that's why I didn't add it. But that's arguable, so let me know what you prefer.

app/src/androidTest/java/ch/epfl/qedit/treasurehunt/TreasureHuntActivityTest.java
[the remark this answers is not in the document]
nbaehler: Oh OK, I see. What a pity! Thanks for your explanation.

app/src/main/java/ch/epfl/qedit/model/Question.java
nbaehler: No, let's keep it that way.

app/src/main/java/ch/epfl/qedit/model/Quiz.java
```diff
@@ -144,14 +138,6 @@ private void checkState() {
...
             }
         }
-
-        /** Checks that the question has the same treasure-hunt settings as the quiz */
-        private void checkQuestion(Question question) {
-            boolean questionTreasureHunt = question.getLocation() != null;
-            if (questionTreasureHunt != treasureHunt)
-                throw new IllegalArgumentException(
-                        "Cannot mix treasure-hunt and non treasure-hunt questions");
-        }
```
nathan96g, line 154: Why do we get ride of this check ?

nathan96g: we could check radius != -1 no ?

app/src/main/java/ch/epfl/qedit/model/Question.java
nathan96g: I agree, this is fine

app/src/main/java/ch/epfl/qedit/model/Quiz.java
brunnerant: I assumed that nobody would create invalid quizzes. But you are right, checking the radius might make the whole thing a bit safer, although not perfect.

Issue #170, opened 2020-05-19
Improve the list edit view by putting the extra actions in a popup menu


brunnerant:
The list edit view was meant to be modular, but it has started to limit us, so the goal of this task is to modularize it even more. It should add a popup menu accessible with three dots for each element of the list view, with a customizable list of buttons.

PR #172, opened 2020-05-19
Improve the list edit view by putting the extra actions in a popup menu

brunnerant:
See #170 for an explanation of what this PR does.


brunnerant:
I have decided to also link the different activities together in this PR, because it was not such a big amount of work. It means that quizzes can now be answered from the quiz list, and treasure hunt quizzes are also dispatched correctly.

app/src/main/java/ch/epfl/qedit/backend/auth/MockAuthService.java
```diff
@@ -34,10 +34,10 @@ public IdlingResource getIdlingResource() {
         return idlingResource;
     }
 
-    /** Simply make the current thread wait 2 second */
-    private static void wait2second() {
+    /** Simply make the current thread wait 0.5 seconds, to do as if the request takes time */
+    private static void fakeWait() {
         try {
-            Thread.sleep(2000);
+            Thread.sleep(500);
```
nbaehler, line 40: You do this to speed up our tests, am I right?

app/src/main/java/ch/epfl/qedit/view/util/ListEditView.java
```diff
@@ -79,61 +81,57 @@
...
-            }
+        // This is called when an item in the popup menu is clicked
+        public boolean onMenuItemClick(MenuItem item) {
+            int id = item.getItemId();
+
+            // We pass as an event the index of the item in the popup menu
+            adapter.notifyItem(getAdapterPosition(), id);
 
-            adapter.notifyItem(adapter.selectedQuestion, EventType.Select);
+            return true; // the even was handled
```
nbaehler, line 134: typo: event

app/src/main/java/ch/epfl/qedit/backend/auth/MockAuthService.java
brunnerant: Yes, exactly. And also to speed up the demonstration that will be using the mocks.

app/src/main/java/ch/epfl/qedit/view/util/ListEditView.java
brunnerant: Thanks, I will will fix it.

brunnerant:
Yeah, I felt like it would make more sense to put it there. And HomeQuizListFragment was starting to be pretty big, so it was also a good move.

Issue #171, opened 2020-05-19
Debug CI check problem on PR#155 that prohibits me from merging.


nbaehler:
The test `testThatActivityIsDisplayed` in `EditMapsActivityTest` is failing on the CI, hence I cannot merge into master PR#155.

Issue #153, opened 2020-05-14
Add two ConfirmationDialogs to the EditQuizActivity and a some visual elements


nbaehler:
One Dialog will ask the user if he really wants to delete question and the other one will notify the user that he can't build an empty quiz.
Additionally I will add a vertical bar the separates the overview from the preview.

Finally I even needed three Dialogs:
1. for deleting a question
2. for quitting without saving
3. for stopping the edition mode and saving the changes made to the quiz.

PR #173, opened 2020-05-20
This PR is only for testing if activity labels can be checked on the CI

nbaehler:
Linked to the PR #155 and the issue #171. Here I try if the test only fails on my branch or if the test generally on the CI. If so I have no idea why @cjordan7 was able to merge his PR that introduced to test. The checks should have failed then as well but we'll see ...


nbaehler:
Well, this works on a branch freshly branched off from master. So I will have to find the problem on my branch addressing #153.

PR #177, opened 2020-05-21
Contains code from #155

nbaehler:
Add all the changes from the somewhat corrupted #155.


nbaehler:
This PR wasn't that useful finally. I had created it because I didn't know and I do not remember it to be it that way but the Ci checks are apparently only launched when the branch is up-to-date with master. So basically I thought that my reverting, resetting had broken the branch entirely which is why I created this duplicate in the first place. Finally managed to find the bug on the original PR #155 so I can close this one without merging.

Issue #178, opened 2020-05-21
Debug missing progress bar in home


nbaehler:
Currently the progress bar which should be shown when loading from the database is not shown in the `HomeQuizListFragment`.

brunnerant:
In the end, this issue took a lot more time than planned. This is because while debugging the progress bar, I discovered some other bugs in the app. All of them were related to concurrency issues with the asynchronous services such as the database and the location. I managed to debug all of that and fix a lot of small issues overall, so that's great.

PR #179, opened 2020-05-21
Debug the missing progress bar in HomeQuizListFragment

nbaehler:
For more details go to #178.

### DIFF
--- a/app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
+++ b/app/src/main/java/ch/epfl/qedit/view/home/HomeQuizListFragment.java
@@ -87,6 +87,7 @@ public class HomeQuizListFragment extends Fragment
 
         // The progress bar is needed while waiting from the database
         progressBar = view.findViewById(R.id.quiz_loading);
+        progressBar.setVisibility(GONE);
 
         // Instantiate Handler and the DatabaseService
         db = DatabaseFactory.getInstance();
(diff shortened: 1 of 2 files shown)


brunnerant:
I managed to fix the issues with the progress bar, but it created issues in the tests, so I have to investigate that part now...

brunnerant:
I finally found where the errors in the tests come from ! The problem was coming from the factory classes that instantiate the database and location services. The method `setInstance` was not properly setting the singleton, so the tests and the app would not use the same service. This was because the methods in those two classes are a bit special in that they take closures as argument, because the services need the context of the app to be instantiated... Anyway, now it should be working, except some minor things that I might have forgotten.

README.md
```diff
@@ -7,6 +7,10 @@ the context of a project at EPFL.
 [![Test Coverage](https://api.codeclimate.com/v1/badges/97269ca5f086f0c8ed40/test_coverage)](https://codeclimate.com/github/brunnerant/sdp-project/test_coverage)
 [![Maintainability](https://api.codeclimate.com/v1/badges/97269ca5f086f0c8ed40/maintainability)](https://codeclimate.com/github/brunnerant/sdp-project/maintainability)
 
+## Wiki
+We have a Wiki were we keep track of some of our design choices, and were we collect some useful information about android
```
cjordan7, line 11: It should be `where` instead of `were`

brunnerant: Thank you for your sharp eye ;), I will fix it.

PR #182, opened 2020-05-21
Sprint11/issue#180

alexisc99:
Had to change qr code generator fragment as it displayed blurry qr codes, now it is fine


brunnerant:
Looks good, but there are a few things that we need to improve before merging. We should make the QR code as big as possible if we want the users to scan it. Maybe we should launch it in a dialog ? Also, I noticed the the home activity doesn't have a vertical scroll bar, so we cannot reach the QR code when the phone is in landscape mode. Then, I think that the QR code should not stay on the home page forever, because it feels weird. Once the user has shared the quiz he probably doesn't want to have the QR code on the home page anymore.

nathan96g (approved):

cjordan7 (approved):

PR #184, opened 2020-05-22
Send new quizzes to database

### LINKED ISSUE #181: Send edited quizzes to database in home
Send the quizzes to the database to make the new or the modified quiz persistent.

nbaehler:
When finishing editing, send resulting quiz to the database. For now we can't update existing quizzes, only upload new ones.

Then use a cache for all loaded quizzes so that we do not have to load quizzes several times. This implementation will change when the general cache will be available.

While writing tests I found a way of implementing some tests that I wanted to do even before this PR, so added them here.


brunnerant:
I noticed another important bug: if the screen is rotated while creating a quiz, all questions are lost, which is pretty annoying. We have to fix it...

brunnerant:
I have to debug why the quiz upload doesn't work. It could come from the `FirebaseDBService` or from the application logic, so I will have to dig into it.

app/src/main/java/ch/epfl/qedit/backend/database/Util.java
```diff
@@ -321,6 +328,7 @@ private static String getBestLanguage(List<String> languages, Context context) {
             CompletableFuture<Void> future = new CompletableFuture<>();
             Map<String, Object> doc = questions.get(i).toMap();
             doc.put("index", i);
+            final int i_ = i;
```
cjordan7, line 331: This has to be removed

cjordan7 (approved):

app/src/main/java/ch/epfl/qedit/backend/database/Util.java
brunnerant: Yes, it does. My bad, I used it for debugging, and I forgot to remove it ;).

nathan96g (commented):
LGTM, (this review has been made by nbaehler, but he could not review this pr because he was the one who created it)

nathan96g (approved):

cjordan7 (approved):

PR #188, opened 2020-06-01
Cache data for offline experience

brunnerant:
This PR will add a decorator database service that caches requests to the real database so that it can be accessed faster, and potentially without internet connection. We will first focus on caching users, so that the login can be done offline, and then we will tackle storing quizzes.

### DIFF
--- a/app/src/main/java/ch/epfl/qedit/backend/database/CacheDBService.java
+++ b/app/src/main/java/ch/epfl/qedit/backend/database/CacheDBService.java
@@ -1,26 +1,30 @@
 package ch.epfl.qedit.backend.database;
 
+import android.content.Context;
+import ch.epfl.qedit.model.Quiz;
+import ch.epfl.qedit.model.StringPool;
+import ch.epfl.qedit.model.User;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
-import ch.epfl.qedit.model.Quiz;
-import ch.epfl.qedit.model.StringPool;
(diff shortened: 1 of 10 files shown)


app/src/main/java/ch/epfl/qedit/backend/database/CacheDBService.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Override
+    public CompletableFuture<StringPool> getQuizStringPool(String quizId, String language) {
+        return retrieve(
+                new File(poolCacheDir, quizId + "-" + language),
+                () -> dbService.getQuizStringPool(quizId, language));
+    }
+
+    @Override
+    public CompletableFuture<String> uploadQuiz(Quiz quiz, StringPool stringPool) {
```
nbaehler, line 81: So then one would store the quiz for instance manually using `storeInCache(...)` before calling this method that uploads the quiz to the real database. This allows more flexibility when deciding what is going to be cached and what not. At first I thought that we might always put the quiz in the cache here as well before uploading to the real database. But as I said this would be less flexible. Did I understand your intentions correctly?

brunnerant: I didn't plan on manually storing the quizzes in the cache, because it would spread the logic to other places in the code. The problem that I though of is that if you upload a quiz and then share it using a QR code, you need the quiz to be loaded in the database. That's why I always upload it for now.
So indeed, what you said would be the only solution to our problem...

PR #190, opened 2020-06-04
 Add method to retrieve the logged-in user in the database services

brunnerant:
This PR will add a method in the database services to retrieve the currently logged-in users. This will allow to make the current implementation of `StartActivity` testable, by mocking the fact that a user might already be logged in.
It will also add log out facilities to the authentication services.


brunnerant:
Nothing is perfect, but it is probably not garbage either ;).

PR #191, opened 2020-06-05
Add a Settings activity

Anthony-Iozzia:


brunnerant:
Do we need a "save" button ? Looking at the code, it just goes back to the home activity.
And do we need to load the user from the database when saving ? Can we not just go back to the home activity, where the user was previously loaded ?

Anthony-Iozzia:
> Do we need a "save" button ? Looking at the code, it just goes back to the home activity.
> And do we need to load the user from the database when saving ? Can we not just go back to the home activity, where the user was previously loaded ?

I tried with a simple "return" arrow at the top left (go to parent) but that way, the HomeActivity is not refreshed and the language is not changed... The save button was all I could think of.

